Issue #133, opened 2017-11-04
Global Navigation README


PabloMorenoVera:
The Global Navigation README is already in spanish, it must be in english.

Issue #132, opened 2017-11-04
Autopark README


PabloMorenoVera:
The autopark README is in spanish.

PabloMorenoVera:
This issue is fixed with the pull request i have created.

PR #134, opened 2017-11-06
Fix #123 and #133 issue: Translation of autopark and global navigation READMEs

### LINKED ISSUE #123: DroneCatMouse: add a viewer in GUI for processed images
Extend the GUI to include not only the raw images from the cat's camera, but also another frame to show the processed images, like the existing for ColorTuner.py tool:
colortuner

It will help the students to debug the perception side of their soluctions to the exercise.

PabloMorenoVera:
The autopark and global_navigation READMEs have been translate into english. These updates of the READMEs fix the #132 and #133 issues
Also update the drone_cat_mouse and vacuum_cleaner READMEs. These updates fix the #139 issue

### DIFF
--- a/src/autopark/README.md
+++ b/src/autopark/README.md
@@ -1,23 +1,24 @@
+
 # Autopark
 
-# Práctica Autopark
+# Autopark Practice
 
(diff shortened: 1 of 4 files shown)


fqez:
thanks!

Issue #160, opened 2018-01-12
Update Visual_Lander practice


pmvera:
This practice must be update with a proper GUI.

Issue #159, opened 2018-01-12
Add Vacuum_Practice to JdeRobot practices


PabloMorenoVera:
This practice must be added to the practices in JdeRbot Academy.

fqez:
I'm not closing this because this is not the final version and it's not working. This will be available soon with @ilope236 work.

Issue #115, opened 2017-03-01
Update Practices to use JdeRobotComm


aitormf:
We need update practices to use JdeRobotComm to be able to use ROS messages:

- [x] 3d_reconstruction
- [ ] bump_and_go_py
- [x] color_filter
- [ ] drone_cat_mouse
- [x] follow_line
- [x] follow_road
- [x] follow_turtlebot
- [x] global_navigation
- [ ] labyrinth_escape
- [x] obstacle_avoidance
- [x] position_control
- [x] rescue_people
- [ ] stop_practice
- [x] vaccum_practice
- [x] visual_lander

jmplaza:
This issue is outdated and no longer relevant as the connecting interfaces between the robot drivers and the exercise templates are not ICE anymore but ROS, and only ROS. In the new design, the templates for all the RoboticsAcademy exercises are now ROS nodes.

PR #161, opened 2018-01-17
Update Visual_lander and add Vacuum_practice

PabloMorenoVera:
This pull_request fix the #160 and #159 issues and finish the #115 issue.


fqez:
Hi @PabloMorenoVera 

thanks for the job! The visualLander practice works fine, but the vaccum does not. It doesn't matter because @ilope236 is working on a imporved version, so I will merge this anyway.

thanks a lot!

Issue #163, opened 2018-01-29
Improve 3Dreconstruction exercise to use 3DViewerJS


jmplaza:
This is also related with issue 1015 at JdeRobot's repo. Maybe the code of the academic node should be updated to use the new 3DViewerJS and the new updated interface for 3D points visualization.
Watch out, the #158 should be merged before this, in order to work with the last (and best) release.

PR #197, opened 2018-04-09
Issue #163: Improve 3Dreconstruction exercise to use 3DViewer-web

rperezgonz:
Improve 3DReconstruction to connect with 3DViewer-web. Now, 3DReconstruction send a buffer points to the 3DViewer when requested.

3DReconstruction is executed as before, I have only modified the way in which points and segments are sending, by creating a new python module called bufferPoint.py


aitormf:
remove solution and this PR can be merged

rperezgonz:
The solution has already been removed. I don't had seen that you had already been changed README file and for that I have changed it, but I've already put yours back

Issue #264, opened 2018-12-21
New exercise: follow_line_turtlebot


jmvega:
A new exercise is going to be developed. The exercise "follow_line_turtlebot" will use a ROS node and the Gazebo Turtlebot model or the real Turtlebot.

jmvega:
The exercise also needs to include the ROS nodes configuration in the .yml file. This way, the student only has to specify in this file whether the Python script is going to be connected to the real Kobuki or the simulated one.

PR #266, opened 2019-01-23
[Issue #264] New exercise follow line with real and simulated Kobuki

jmvega:
The exercise is finished.


aitormf:
Can you remove solution to accept it?

PR #288, opened 2019-03-23
car_junction exercise to ROS migration

kurshakuz:


cawadall:
Well done @kurshakuz ! I've checked your patch and everything works fine for me with the proposed solution. Before we can merge it with the master branch, please delete the previous node ('/old' directory) and recover the configuration file ('stop_conf.yml') with the new ROS topics, since we need a distributed and modular structure so students are able to modify and add new nodes in a simply and comfortable way. You can find an example of reading and parsing the configuration file in [here](https://github.com/JdeRobot/RoboticsAcademy/blob/master/exercises/laser_loc/laser_loc.py), lines 431-464.

Keep working like that! Regards.

kurshakuz:
Hi, @cawadall! I have added the configuration file, removed the /old directory, and updated the readme and initialization script. Hope everything is fine now. 

And thank you very much for the detailed feedback! The purpose of the configuration file was ambiguous for me, but now everything is clear.

PR #297, opened 2019-04-02
Fix ROS version of auto_park exercise

Kareem-Emad:
Addressed the following issues:

- Changing Configuration Yaml file to ROS gives error when running the exercise.
- Opening parking world using gazebo does not load or initialize Gazebo node for receiving commands and publishing  sensor data.
- Topic names does not match published topic names.
- Errors arise when listening to laser data from mapWidget.

### DIFF
--- a/exercises/autopark/README.md
+++ b/exercises/autopark/README.md
@@ -5,13 +5,11 @@ The goal of this practice is to implement the logic of a navigation algorithm fo
 ## How to run
 To launch the example, follow the steps below:
 1. Run Gazebo simulator:
-    * Execution without seeing the world: 
-       `$ gzserver autopark.world`
-    * Execution watching the world: 
+    * Execution watching the world:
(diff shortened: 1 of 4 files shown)


kurshakuz:
Hi @Kareem-Emad! I've ran your edition and faced the following problems:
- Lidar data is not showing anything in GUI.
- autopark GUI is not closing properly.
- referee.py is using old conf files and is trying to connect to ICE interfaces.

I've already solved those problems in my PR, so feel free to look at it as well.

Kareem-Emad:
Hi @kurshakuz , I only addressed the problems that I mentioned above in this PR ,I was going to push more updates on this ,but after you added your PR I stopped as you already did most of what I'm working on. so I think your PR will more eligible if  you integrated comm package, that way we will have only one implementation for all listeners/publishers in jdeRobot/base and we could update them from there.

PR #298, opened 2019-04-03
Autopark to ROS migration

kurshakuz:
Hello!

I've finally fixed the problem with lidar visaulization in GUI, and completely moved this exercise to ROS. Everything was done in compliance with RobotcsAcademy standards (all previous suggestions were followed) and hopefully can be merged. The only problem rn is that the world itself referenced in roslaunch file is not complete. I'm not sure how can I edit files in root JdeRobot directory, so will be glad if you can help me with it. The only thing that is required there is to add some more cars so car can park between them.

If there's something wrong here please let me know, I'll fix it asap.


Kareem-Emad:
I think there is already an implementation in the 'comm' python package for pose/laser listeners and motor publisher,it will be better if we re-use those instead of implementing them in every exercise.

Note: you can check it out in ``/opt/jderobot/lib/python2.7/comm/``

kurshakuz:
@Kareem-Emad thank you very much for pointing that out! I was just referring to the https://github.com/JdeRobot/RoboticsAcademy/pull/288#issuecomment-476574409 comment and thought that it is the structure we have to follow, but this one actually makes more sense. I have updated both the autopark and referee to use these packages

cawadall:
The 'comm' python package is included in the JdeRobot package. The reason for not using the 'comm' interfaces in some exercises is that we wanted to build an infrastructure for the exercise independent of the root package, in such a way that the volume of packages to be installed for potential users is reduced.

However, that has been done only in those exercises in which a Jupyter template has been made. Therefore, for the moment it is not necessary to make this exercise independent of that package.

Everything works fine for me, good job @kurshakuz !

Issue #364, opened 2019-12-11
obstacle_avoidance exercise not the same as given in website


SakshayMahna:
The obstacle_avoidance exercise given in the [website](https://jderobot.github.io/RoboticsAcademy/portfolio/obstacle_avoidance/) is based on Formula1 cars, whereas the actual exercise consists of TurtleBot robot.

jmplaza:
Solved in #365

PR #365, opened 2019-12-11
Changes to obstacle_avoidance

SakshayMahna:
This is in reference to issue #364.
Added the functionality to choose from TurtleBot Robot or Formula1 Car. Also, fixed the error related to target and vectors in map widget and added a launch folder to allow for easy instructions.


SakshayMahna:
Added HINTS.md and THEORY.md as well which is in reference to issue #366

SakshayMahna:
Updated the Webpage of the exercise

PR #383, opened 2020-01-06
Obstacle avoidance

mohamedsayed18:
there was an error "MainWindow has no attribute setCamera"
so I just uncommented the lines in the gui/GUI.py


Korusuke:
I faced the same error but instead of uncommenting `setCamera`, I removed the function call from `obstacle_avoidance_f1.py`. 
Since this exercise doesn't need a camera anymore shouldn't we just remove all the code related to it?

Also on a side note, the instructions page mentions to edit `MyAlgorithm.py` but the `_f1` config uses `MyAlgorithm_f1.py`.

Issue #393, opened 2020-01-14
Images not visible on Vacuum Cleaner with Localization


SakshayMahna:
There is a typo error in the markdown file of the exercise.

jmplaza:
Solved in #394

PR #394, opened 2020-01-14
Update vacuum_cleaner_loc.md

SakshayMahna:
This is in reference to issue #393. Images will now be visible.


SakshayMahna:
The directory name and title have been changed as well.

SakshayMahna:
Added the Table of Content as well

PR #410, opened 2020-01-19
Review index. Some images are changed in the cover and exercise index.

igarag:
You can't put up a video gallery right now. We'll keep investigating. We're leaving the full video in place temporarily.


igarag:
fix #410

PR #484, opened 2020-03-21
fix security issues

fqez:


igarag:
Security alert solved.

PR #499, opened 2020-03-27
car_junction to work on ROS Melodic

kurshakuz:


kurshakuz:
Solved in #537 and #539

PR #555, opened 2020-06-25
Update broken URL in Installation Guide

joeljosephjin:
update "Get FastRTPS and FastCDR" url

https://github.com/eProsima/Fast-RTPS-docs/issues/171

As you can see here, the link has been changed


pariaspe:
Thanks @joeljosephjin

PR #569, opened 2020-08-03
Load and Save local code - web templates

davidrol6:


exercises/follow_line/web-template/assets/ws_code.js
```diff
@@ -53,12 +72,32 @@ function stopCode(){
     websocket_code.send(stop_code);
 }
 
+// Function to save the code locally
+function saveCodeLocally(){
+	var python_code = editor.getValue();
+	var blob = new Blob([python_code], {type: "text/plain; charset=utf-8"});
+	if (window.navigator.msSaveOROpenBlob)
+		window.navigator.msSaveOrOpenBlob(blob, "follow_line.py");
```
SakshayMahna, line 80: The file name to which we are saving should be some common name, for instance: `exercise.py` or `academy.py`

exercises/follow_line/web-template/assets/ws_code.js
```diff
@@ -53,12 +72,32 @@ function stopCode(){
...
+		a.click()
+		setTimeout(function(){
+			document.body.removeChild(a);
+			window.URL.revokeObjectURL(url);
+		}, 0);
+	}
+}
+
 // Function to save the code
 function saveCode(){
```
SakshayMahna, line 96: I think we can comment the code that uses websocket connection to save and load.

jmplaza:
Those instructions will be useful in the online usage of Academy, through a webserver, so maybe we just may comment them for now. They are not needed in the offline usage of Academy, but we will use them in the online usage.

SakshayMahna:
This is in reference to issue #564

PR #598, opened 2020-09-14
Tweaked manager

davidrol6:
One of the commands gave an error which delayed the execution and caused the simulation to start later, raising a connection error.


pariaspe:
Solved in #597

Issue #600, opened 2020-09-17
Follow line web-template error


NayOoLwin5:
Hi,
While executing the host.py file, got the following error

```
Traceback (most recent call last):
  File "host.py", line 287, in <module>
    server = Template()
  File "host.py", line 38, in __init__
    self.host = sys.argv[1]
IndexError: list index out of range
```
I have tried with all these python versions : 3.8.5 , 3.6.9  and 2.7.17. Still the same error for all.

SakshayMahna:
Hello!
While executing the host.py file, you need to enter the IP address at which the web templates will publish their interface. Since, you may be running this in a Docker container, it is the IP address with which your Docker communicates with your native machine.

On a Windows Machine the IP address is something like 192.168.x.x. So the command would be

```bash
python host.py 192.168.x.x
```

However, if your are on a Linux Machine, the IP address would be something like 127.0.0.x. So the command would be

```bash
python host.py 127.0.0.x
```

**Since, we are using ROS Melodic, all the software has been written in Python 2.7**

PR #601, opened 2020-09-18
Follow line web-template README.md modified 

NayOoLwin5:
This is in reference to #600

### DIFF
--- a/exercises/follow_line/web-template/README.md
+++ b/exercises/follow_line/web-template/README.md
@@ -2,22 +2,50 @@
 
 ## How to launch the exercise?
 
-- Launch the Gazebo simulation
+- There are two ways to run the exercise using web-template.Either one way is fine.
+
+      - Run the exercise with docker container
+      - Run it without container
+
+## Run with docker container
+
+- Jderobot already has created docker image available from [docker hub here](https://hub.docker.com/r/jderobot/robotics-academy/).
+-Follow the instructions line by line to run the exercise.
+
+## Run without docker container
+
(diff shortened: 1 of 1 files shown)


SakshayMahna (commented):
Hello Nay!
One thing you can do in order to completely improve this README.

Possibly, you must be following instructions from the Docker-Hub page. [This one](https://hub.docker.com/r/jderobot/robotics-academy)

Try adding all the instructions from the Docker-Hub README to this README. In this way, we may have the complete instructions on how to run this exercise using Docker.

SakshayMahna:
I also just realized in case of Windows the IP address passed with the `python host.py` is **not** of the form `192.168.x.x`.

In place of that it is `172.17.0.2` in my case. This is present in the file `/etc/hosts` in the Docker container.

Capture

NayOoLwin5:
Hi @SakshayMahna,
I updated the README file. Please have a look at it.

I wasn't using docker container at first . But I still needed to pass the IP address 127.0.0.53 on my native linux for the established connection.
Running `cat /etc/hosts` on linux without docker doesn't show any IP address. Therefore I included `cat /etc/resolv.conf` in README file which also shows the wanted IP address .

Later,I tested docker container on linux machine on which the IP address shown with the command `cat /etc/hosts` is same 172.17.0.2 as yours.
Screenshot from 2020-09-20 07-27-54
It seems it is same on any machine if you are using the docker container

SakshayMahna:
Great! I am merging the PR now.

PR #612, opened 2020-10-06
Instructions added for CustomRobots model

### LINKED ISSUE #606: F1 position needs to be changed
Inside f1_circuit.world file of follow line web-template directory,the pose is **53.462 -10.734 0.004 0 0 -1.57** which shows the following 
Screenshot from 2020-09-25 05-24-36
It needs to be changed to **53.462 -10.734 0.004 0 0 0** for correct starting position.

NayOoLwin5:
fixes #606 

Directing native installation users to use CustomRobots f1 model. Thus, there will be no bugs for running follow_line web-template without using docker.


SakshayMahna:
Really useful set of instructions!

PR #632, opened 2020-10-30
No module named error fixed

NayOoLwin5:
No module named error appears in the console while using the web-templates. It is changed to correct form.


SakshayMahna:
Hello Nay!

I think you are still using the 0.2.2 version(tag) of Docker Image. I have recently updated the tag to 0.2.3. In this version, even though we import the HAL and GUI, we are not able to use their functions outside the `while True` loop(in the sequential code area)

The 0.2.3 version, fixes this issue. In this version, the student needs to `import from HAL` rather than `import from hal`.

PR #665, opened 2020-12-14
[ColorFilter exercise] WebRTC driver for web template of color filter…

vmartinezf:
Added web template of color filter

This issue has been the result of the 654


dvalladaresv:
It has already been published in the issue-654

PR #698, opened 2021-01-09
Updated Follow Line new Frontend Robotitcs Academy offline webtemplates

martaquintana:
I have added new frontend style to the exercise follow line, I hope you like! 

Regards


SakshayMahna:
Hello!
The exercise interface looks very good now! There seem to be just two small problems:

- Could you adjust the `src` attribute of the `<iframe>` tag, to use variable websocket address? There is a file in assets by the name of `websocket_address.js` which defines a variable `websocket_address`. Use this variable as the attribute(by string concatenation of port number), this may allow users with different Docker IP addresses to connect.
- The pre-written code in the ACE Text Editor, seems a bit mis-indented. I think this can be adjusted by removing the tabs and spaces in the relevant HTML portion.

martaquintana:
Hello!  
Ok! I'm going to see what  you say about the editor,  but about the iframe source it would be like this:

src='{web socket_address}:8080' 

or with double quotes??
Thank you

SakshayMahna:
Sorry, I don't really know about this notation: '{websocket_address}'

As far as I know, we can use Javascript variables in HTML, in a script. Something like, `document.getElementById("iframeGzweb").setAttribute("src", "http://" + websocket_address + ":8080")`, as illustrated [here](https://www.w3schools.com/jsref/met_element_setattribute.asp)

Will the notation you mentioned, accomplish the same result?

martaquintana:
It's true I was confusing it with unibotics-webserver,  I just solved it. Check it
Thank you!!

SakshayMahna:
Great!

PR #759, opened 2021-03-11
gzclient + vnc + novnc for follow_line(fixes #740)

### LINKED ISSUE #740: Test gzclient + vnc + novnc as a replacement of gzweb in RADI
Current RoboticsAcademyDockerImage (RADI) includes gzserver and gzweb servers. From the web based exercise template, that is, the exercise webpage gzweb is used. The browser connects to those servers running inside RADI. The Gazebo simulated scenario is shown in the browser, outside RADI, this way.

Another interesting approach is running gzserver + [gzclient + vnc display server + apache] inside the RADI and novnc viewer in the browser.

Suhas-G:
The pull request fixes #740 
The intention is to allow running gzclient inside docker container and viewing it in host using [novnc](https://github.com/novnc/noVNC). It also has changes for documentation and renaming host.py to exercise.py
Steps:
1. Installed x11vnc inside container
2. Exposed a port 5900 for the container to communicate with x11vnc
3. Started gzclient from container terminal, with appropriate height and width depending on host's browser.
4. Start x11vnc from container terminal.

This is my first PR in this repository. So please review and let me know for any modifications to be done.


SakshayMahna:
Great! I'll check the changes and suggest any feedback.

Suhas-G:
I have resolved merge conflicts.

SakshayMahna:
Hello!
So I went through the PR. I followed the mentioned instructions and the VNC variant works great!

But there is one small thing that we need to work upon. As per the instructions, after starting the Docker container, we need to **download and run the noVNC client from our local machine manually**. This is an **additional step** for the user to do after running the Docker container which should **not be present**.

The idea with Web Templates version of Robotics Academy is the user has to go through minimal installation. The entire workflow, should be like: 

1. Download Docker
2. Run the Docker container
3. Start the Web Browser

And the exercise should start working. For now, viewing the simulation in a separate tab is good. But the extra steps, to start the client should be avoided.

Let's look into it, how this can be resolved.

Suhas-G:
Hi. Thanks for the feedback. I'll try to move novnc to docker itself.

scripts/Dockerfile
```diff
@@ -211,6 +211,16 @@ RUN apt-get install -y nodejs
 RUN git clone https://github.com/osrf/gzweb.git
 RUN cd gzweb && git checkout gzweb_1.4.0 && npm run deploy --- -m local
 
+# VNC
+RUN apt-get -y install x11vnc
+
+RUN git clone https://github.com/novnc/noVNC.git &&\
```
Suhas-G, line 217: noVNC is officially available through snap. But installing snap also requires systemctl, which is not available in this image. So copying the files from git and checking out the latest release.

Suhas-G:
I have updated the pull request with changes for noVNC within docker. Now the use can directly open the gazebo simulation in browser as updated in follow_line.md.

@SakshayMahna : Let me know if this is fine

SakshayMahna:
Great work!

Sorry, I didn't notice this earlier but this is a small change: Since we are going to replace gzweb with the vnc viewer. In the photo below, Gzweb is displayed in the tile at right bottom side.

Web Template

The above is the view of [this file](https://github.com/JdeRobot/RoboticsAcademy/blob/master/exercises/follow_line/web-template/exercise.html). This file contains at line 138 the implementation of `iframe` that displays gzweb, `document.getElementById("iframeGzweb").setAttribute("src", "http://" + websocket_address + ":8080")`. We should change the `:8080` part to `:6080/vnc.html` or `:6080/vnc_lite.html`. Probably, the one that connects without the CONNECT dialog button. 

**Note**: Keep the `websocket_address` variable as it is, it is useful when we do not want to use Docker connection on localhost.

scripts/Dockerfile
SakshayMahna: I found this [another way](https://www.server-world.info/en/note?os=Ubuntu_18.04&p=desktop&f=7) as well, just for reference.

This installs novnc through `apt`, and then we can run the novnc client using `websockify` command.

Suhas-G: Yep. This seems to be a better way, instead of depending on a particular tag in git.

Suhas-G:
I tried this, but computing width and height of iframe which is not loaded is trickier. (I'm sending the heigth and width to the docker so that gzclient can open with that configuration. Otherwise it opens in some default window size which does not span the whole view. )
image
With some calculation I can pass width and height, but its not completely correct and might change based on screen size (probably because of padding / borders). Users can scroll around, but its still not that user friendly. Any suggestions?

jmplaza:
Hi @Suhas-G , 

yes we may start the gzclient window inside docker with a certain size [https://answers.gazebosim.org//question/12804/is-there-a-way-to-start-gazebo-with-a-smaller-window-size/](https://answers.gazebosim.org//question/12804/is-there-a-way-to-start-gazebo-with-a-smaller-window-size/)

Suhas-G:
@jmplaza . Yes, I'm following the same method. So to get the width and height, if its a new tab it is pretty easy. But for an iframe element because width and height are calculated depending on the layout, I don't have this info at page load time when exercise command is sent [here](https://github.com/JdeRobot/RoboticsAcademy/blob/b9edad71de4ea2e5ebfa771ed23c6a328128631c/exercises/follow_line/web-template/assets/launcher.js#L4) as iframe is not loaded yet.

I can calculate manually to an extent (based on the style property of iframe), but it might not be the exact value available.

Would it be okay to make the Gazebo button open a new tab instead of showing the iframe?

jmplaza:
Uhm, opening Gazebo display in a separate tab would be bad for usability... 

I see four different parts here:
1.- DISPLAY size in the Xserver inside docker. We can set this to certain values, can't we?
2.- The position and size of gzclient window inside that DISPLAY. This can be set at will too, just using the .gazebo/gui.ini settings
3.- The x11vnc server. Does it provides the whole DISPLAY(a) or can it provide only certain window inside that DISPLAY(b)?
4.- The position and size of the iframe inside the browser webpage, it runs the novnc client. This for sure will depend on the user's computer-monitor and it may change at run time if the user changes the browser window...

The sizes of (2)-(3a)-(3b) and (4) may not fit. Hopefully the "scale" configuration of  novnc will help on this. [https://github.com/novnc/noVNC/issues/869](https://github.com/novnc/noVNC/issues/869). Let's explore it.

Can we see only a specific window in the DISPLAY into the novnc?

Suhas-G:
Hello,
1 -  Yes, we can set the xserver display config. Right now it seems to be taken from https://xpra.org/xorg.conf
    And its defaulting to 8192x4096
3 - We can show only a particular window based on its id. (Have to check more if there's way to identify based on process name)

The scale option isn't present as part of command line argument or the settings in novnc display. The referenced issue  869 requires code changes in JS file I think. I'll look into it some more.

SakshayMahna:
Hello! 

I have some findings which might be useful for the task:

1. The "scale" configuration of noVNC mentioned by @jmplaza, can be set as GET query parameters of the `vnc.html` page, as mentioned in their [documentation](https://github.com/novnc/noVNC/blob/master/docs/EMBEDDING.md). The configuration which would be useful for us would be: `?autoconnect=1&resize=scale`.

2. One way we could fit the iframe would be to set the initial size of Gazebo equal to the XWindow size, set in `xorg.conf`. I tested the same with a smaller XWindow size of `1024x768`, with the `resize=scale` option. Following is a picture of the same. This method however, may require some more refinement, as Gazebo doesn't fit entirely in the iframe.

Screenshot (6)

Suhas-G:
Hello! Sorry for the delay. I was working on the programming challenges.
Based on @jmplaza and @SakshayMahna comments, 
a. For the resize option, apart from "scale" there is "remote" option as well. This worked better even without changing the Xorg display config.
b. It was possible to show only Gzclient window based on its ID (used xwininfo). But its not really required, the display was still same as when using remote option directly. And also displaying only a particular window introduces its own problem where the menu becomes blacked out.
image

So showing the whole screen with resize = remote and with some calculation for width and height, it can be shown like below.
novnc-0-10

Let me know if this seems ok. Thanks!

jmplaza:
Hi @Suhas-G , @SakshayMahna , interesting findings! Displaying only the gzlient window looks good to me. 

@SakshayMahna maybe that scaling does the best possible fit with that iframe size. So we should use the same aspect ratio for both the iframe and the gzclient window. This way whenever the iframe is resized the system will do the best scaling.

In addition, we should configure gzclient to not showing any vertical side menu at the very beginning. Showing the horizontal menu (File, Edit, Camera, View, Window, Help) will allow the user to open it at will if needed for a while.  Is that feasible?

Suhas-G:
I tried finding a way to close the left hand pane automatically. There is no command line argument for fullscreen mode in gzclient and also it doesn't seem to be configurable. (I didn't find any proper documentation for possible options in gui.ini but a [github search](https://github.com/search?q=org%3Aosrf+getINIProperty&type=code) didn't show anything related to it )

But it might be possible to make it full screen by simulating a F11 keypress from terminal (can be done using a tool called xdotool) or writing some gazebo plugin (I'm not too sure about this yet). Below image is how it looks in fullscreen mode inside iframe
image

SakshayMahna:
Looks great!

The full screen configuration can be set in the world file for the exercise itself. This way, the side panel would be hidden by default.

This configuration can be put in both these files [[1]](https://github.com/JdeRobot/RoboticsAcademy/blob/master/exercises/follow_line/web-template/launch/f1_circuit_headless.world) and [[2]](https://github.com/JdeRobot/RoboticsAcademy/blob/master/exercises/follow_line/web-template/launch/f1_circuit.world), present in `exercises/follow_line/web_template/launch`. In these files under the `world` tag make these changes: add a new `gui` tag with an attribute by the name of `fullscreen=1`. This would become:

```
<world name='default'>
<gui fullscreen=1></gui>
...
```

This may configure the required settings. [Here](http://sdformat.org/spec?ver=1.7&elem=world#world_gui) is the documentation for the same.

Suhas-G:
Ah ok. I didn't consider this angle. I'll try this and update the code. Thanks!

Suhas-G:
I have updated the code with the required changes. I tried the novnc through `apt` as shown [here](https://www.server-world.info/en/note?os=Ubuntu_18.04&p=desktop&f=7) but I think its an older version which doesn't support resize option. So I have continued with the git repo of novnc.

SakshayMahna:
Merged

Suhas-G:
Thanks! Is there any other issue I can take up? Any suggestions?

jmplaza:
Sure @Suhas-G, let's extend this GUI technique to all available exercises in RoboticsAcademy 2.3

Suhas-G:
Sure @jmplaza, I'll do that.

PR #769, opened 2021-03-17
add Connect button on exercises webpages

### LINKED ISSUE #766: Introduce the "Connect button" on all exercise webpages
This way the state of the connection with the underlying exercise.py, which runs inside the RADI, is explicit and clear.

nsoul97:
This is PR fixes #766.
I have added a connect button only for the follow line exercise, to get some feedback before adding the button to the other web template exercises as well.

The button text describes the current status of the connection (Connected/Disconnected). Once you move the cursor over the button, the text describes what will happen if you click the button (Disconnect/Connect). 

When the button was clicked immediately after the page was loaded and before ws_manager had received 2 "Done" messages, I was getting an error ERR_SOCKET_NOT_CONNECTED when trying to create a new websocket (in the declare_code and declare_gui function calls). To solve this problem, if the ws_manager hasn't received the "Done" messages and the button is pressed, an alert message is shown and the user will have to press again the button in order to connect.

I did not place the connection button with the other buttons (Start, Stop, Save, Load, Reset), as I think that their functionality is different and this should be depicted in the UI.

Any feedback will be much appreciated.

### DIFF
--- a/exercises/follow_line/web-template/assets/launcher.js
+++ b/exercises/follow_line/web-template/assets/launcher.js
@@ -6,11 +6,12 @@ ws_manager.onopen = function(event){
     ws_manager.send(JSON.stringify({"command": "open", "exercise": exercise}));
 }
 
+//We must receive 2 "Done" messages before we can connect/disconnect
+var rcv_msgs = 0;
 ws_manager.onmessage = function(event){
     console.log(event.data);
+    rcv_msgs += 1;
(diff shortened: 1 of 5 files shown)


dvalladaresv:
Thank you very much for your PR, it's a very good job!

In the end we have designed a base template to avoid having to put the button in all the exercises.

Any new input is welcome! Thanks again for the help

PR #787, opened 2021-03-27
New buttons: play, pause and reset mouse

### LINKED ISSUE #786: [Drone cat mouse] Pause and Reset buttons to web templates version
**Pause button** will stop the mouse, which will remain hovering at the current position.
**Reset button** will reset the mouse, changing flight mode to landed and moving drone to initial position.

pariaspe:
Closes #786


pariaspe:
Reset mouse is not an instantaneous operation

PR #792, opened 2021-03-29
adding connect/disconnect button

shreyas2000:
button for connecting/disconnecting exercise.


jmplaza:
Hi @shreyas2000 , thanks for this PR. We have just solved the issue in https://github.com/JdeRobot/RoboticsAcademy/pull/791
Nice contribution, anyway. Thanks! The good point is that now all the exercises have this button and RoboticsAcademy users may benefit from it.

PR #805, opened 2021-04-03
follow_turtlebot_v2.3

iamarkaj:
Migrate follow_turtlebot to Robotics Academy 2.3


pariaspe:
Hello @iamarkaj,

Thanks for the PR, some aspects about it:
- Do not use the roomba model, use instead our [turtlebot model](https://github.com/JdeRobot/drones/blob/1477aa1bc4a114410e371e8ff7d8636345a9c7f0/drone_assets/urdf/turtlebot3_waffle_pi_colored_plate.xacro). 
- World file has to be modified, roomba is not needed.
- Launch file has to be modified to spawn the turtlebot. Group tag is not needed, only one drone is spawned.
- YAML file is not needed. No namespaces have to be used.
- HAL file has to be modified. Namespace not needed in DroneWrapper instance. Cat should be renamed as "drone" or similar.
- HTML file has to be also modified. Mouse btns have to be removed. New necessary buttons can be added in future PR.

iamarkaj:
Hi @pariaspe 

Please review the changes especially the dockerfile

Regards,
Arka

pariaspe (approved):

pariaspe:
PR reviewed and merged. Good job @iamarkaj !

PR #809, opened 2021-04-04
Updated home.md

dgvsnl:
Updated the grammatical error on the home page


trunc8:
There is a spelling as well as grammatical error in your changes. Just 'her' -> 'their' would do good.

PR #830, opened 2021-04-07
[RADI] Update README

Jovian-Dsouza:
Update port mapping in the Installation Instructions


pariaspe:
Ports mapped might be revisited in the future, merging anyway.
Thanks @Jovian-Dsouza

Issue #838, opened 2021-04-08
[drones] Clean ColorTuner from exercises


pariaspe:
Since [ColorTuner](https://github.com/JdeRobot/ColorTuner) has been archived and not used in v2.3, ColorTuner references in documentation need to be removed. 
Conf files can be also removed.

RUFFY-369:
@pariaspe  Opened a PR for it

PR #839, opened 2021-04-08
[drones] Fix for Issue #838 

RUFFY-369:
Fix to solve issue #838, cleaning conf files related to color tuner and its references in docs


pariaspe (changes requested):
Hi @RUFFY-369, I'm not sure if color_filter exercise need those YAML files, better to not remove them

pariaspe:
Done in #850

Issue #843, opened 2021-04-09
Monitoring Gazebo Real-Time-Factor in real time and show it on the exercise webpage


jmplaza:
Right now we already monitor Brain frequency and GUI frequency (they both also allow setting their desired frequencies through a slider). Monitoring the "simulator frequency" too may provide information about the computation status of the running exercise. For instance, is the student's computer fast enough to run the exercise? is Gazebo's performance degrading and not reaching the Rea-Time-Factor of 1?

One possible implementation for this is in the exercise.py, just running the "gz" tool ("gz stats" command), processing the results and sending them to the browser through the websocket "ws_code" (the same used for sending the measured brain frequency and the measured GUI frequency).

trunc8:
Hi, please can one of the maintainers look at #844, in which I have tried to solve this issue?

@jmplaza and @SakshayMahna some of the threads in the exercise.py like measuring frequency and tracking real-time-factor run perenially. It might be more efficient to run them at intervals. What is your opinion on this? Would this problem get solved with Python multiprocessing?

Also, Sakshay is already working on the shift to multiprocessing architecture. Can I take more involvement in that task?

SakshayMahna:
Hello!
The measuring threads take up very small part of the computation. Maybe, running it in intervals won't make much difference. 

I already have completed my side of the work on multiprocessing architecture, you can have a look at #727, and suggest any changes or improvements.

jmplaza:
Hi @trunc8 , 

tracking real-time-factor inside exercise.py seems better for me than using Intervals at the web browser. As @SakshayMahna pointed out the take up very little computation. 

Even further, inside the browser the Intervals are dangerous on heavy load situations. In case of measuring frequencies in the browser I would choose Timeouts (creating the next Timeout at the very end of the current Timeout), so they would not accumulate in the case of heavy load situations.

ReyDoran:
This issue was solved on pull request [#1035 ](https://github.com/JdeRobot/RoboticsAcademy/pull/1035)

PR #844, opened 2021-04-10
Display Real time factor in the follow_line's exercise webpage

trunc8:
Solves Issue #843 
image
This is how the final render looks(compare RTF in the table and in Gazebo VNC). I have described my changes via verbose commit messages.

### DIFF
--- a/exercises/follow_line/web-template/assets/ws_code.js
+++ b/exercises/follow_line/web-template/assets/ws_code.js
@@ -63,10 +63,6 @@ function submitCode(){
     var python_code = editor.getValue();
     python_code = "#code\n" + python_code
     
-    // Get the debug level and add header
-	var debug_level = document.querySelector('input[name = "debug"]').value;
-    python_code = "#dbug" + debug_level + python_code
-    
     console.log("Code Sent! Check terminal for more information!");
     websocket_code.send(python_code);
 
--- a/exercises/follow_line/web-template/assets/ws_code.js
+++ b/exercises/follow_line/web-template/assets/ws_code.js
@@ -43,12 +43,22 @@ function declare_code(websocket_address){
 			document.querySelector("#ideal_gui_frequency").value = frequency_message.gui;
 			document.querySelector('#ideal_code_frequency').value = frequency_message.brain;
 		}
+		else if(operation == "#stat"){
+			var stats_message = JSON.parse(source_code.substring(5,));
+			// Parse real time factor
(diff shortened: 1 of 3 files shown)


SakshayMahna:
I will review the PR and let you know the details.

SakshayMahna:
Extracting the RTF from gz stats seems good to go!

But, there is one thing; try including the RTF message in the `frequency_message` dictionary and the sending part to `send_frequency_message` function itself. This would also result in an appropriate change to the `ws_code.js` file. 

This would maintain uniformity in the messages being transferred through the websockets(Single JSON packets containing relevant information)

trunc8:
Cool! That was actually the first version that I had implemented and got working. But the nomenclature seemed off. Naming it as "stats" generalizes if we want to display simulation and real-time in the future. Shall I modify it back as part of the frequency_message?

SakshayMahna:
Since, it's a single simulation based message, we can work with the naming convention for now. On addition of more such simulation elements, the `frequency_message` would be renamed to something like `control_message`.

According to the design, the code websocket has two functions, exchanging the student's code and exchanging the control messages(that's why the control message rename would be appropriate)

trunc8:
Yup, that makes sense. I will include it as part of `frequency_message` for now. What is the organization's preference at such times? Shall I amend the last commit `758184f` and force push the modified commit on my branch `issue-843`? This will keep our master branch's commit history cleaner.
Or shall I simply add a new commit to `issue-843`?
[Reference- Options 1 and 2](https://www.burntfen.com/2015-10-30/how-to-amend-a-commit-on-a-github-pull-request)

SakshayMahna:
You can simply, add a new commit to the branch.

jmplaza:
We may even show the Gazebo RTF continuous measurements in the frontend just in the place of "DebugLevel widget". which may be safely removed. With the enhanced console (based on the vnc display), it makes no sense. It should be implemented by the students on their own code in case of being interested on several levels for debugging messages.

trunc8:
I have removed the debug-level widget and enforced a single JSON message in separate commits. I have verified that the bot completes a lap with the present changes. Requesting one of the maintainers to review the changes.

SakshayMahna:
Great! The changes look good to go!

There is just a slight concern, since the removal of Debug Window would require us to keep both the repository and the RADI image up to date, we may merge these changes after the 0.3.14 RADI image has been released. The release would be updated in just a couple of days.

Issue #806, opened 2021-04-03
[FollowTurtlebot] Migration to Robotics Academy v2.3


iamarkaj:
TODO
- [x] It is currently using roombaROS from [CustomRobots](https://github.com/JdeRobot/CustomRobots), this needs to be fixed because using the same robot will reduce the difference between version 2.3 and version 2.1 (done in #805)
- [x] Update readme, webpage and add clouds (done in #835)
- [x] Update RoboticsAcademy/RoboticsAcademy.html to include follow turtlebot (done in #840)
- [x] More buttons must be added to control the ground robot (done in #849)
- [x] Modify launch file to use turtlebot xacro from drone_assets package instead local duplicated file (done in #857)
- [x] Update developer launching instructions inside RoboticsAcademy/exercises/drone_cat_mouse/web-template/README.md and RoboticsAcademy/exercises/follow_turtlebot/web-template/README.md since it no longer uses gzweb (done in #863)
- [ ] Fix Reset button. Drone flies randomly and crashes sometimes. Maybe due to lat and long. [This](https://github.com/mavlink/mavros/blob/7b7790026f710c2888ff28cc181e46b6b94acdbd/mavros/scripts/mavcmd) may be useful.
- [x] Fix console.print() and print error messages of the student's code
- [ ] Update YT video inside RoboticsAcademy/RoboticsAcademy.html to correctly demonstrate web-template based v2.3 of this exercise

dgvsnl:
I want to contibute to this issue

pariaspe:
Done in #805. RADI 0.3.14 will include exercise dependencies.

Some tasks remaining:
- Add buttons to start, pause and reset turtlebot
- Improve gazebo world to look nicer
- Modify launch file to use turtlebot xacro from drone_assets package instead local duplicated file

Credits to @iamarkaj for the migration

pariaspe:
Some other tasks to do as @iamarkaj said [[1]](https://github.com/JdeRobot/RoboticsAcademy/pull/835#issuecomment-815394833):

1. Update developer launching instructions inside RoboticsAcademy/exercises/drone_cat_mouse/web-template/README.md and RoboticsAcademy/exercises/follow_turtlebot/web-template/README.md since it no longer uses gzweb
2. Update RoboticsAcademy/RoboticsAcademy.html to include follow turtlebot

trunc8:
[...]

PR #849, opened 2021-04-12
add play,pause and reset buttons to control ground turtlebot

iamarkaj:
issue #806

### DIFF
--- a/exercises/follow_turtlebot/web-template/assets/main.css
+++ b/exercises/follow_turtlebot/web-template/assets/main.css
@@ -146,6 +146,21 @@ margin-bottom: 10px;
     border-image-slice: 1;
 }
 
+#turtlebot_stop_btn{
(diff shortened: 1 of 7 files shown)


iamarkaj:
Hi @pariaspe 
please review this

pariaspe:
Thanks @iamarkaj, your PR works well.

Just one comment, previous to migration to v2.3 the turtlebot started to move after launch. This behavior was kept after migration, but since we have buttons now, it would be better to start moving the turtlebot after pushing play button. This also extends to reset button, which resets turtlebot position well but does not stop the robot.

iamarkaj:
Hello @pariaspe the turtlebot does not start automatically now, but for the first time the play button is pressed, it waits for few seconds before playing

pariaspe (approved):
Looks nice, thanks @iamarkaj

Issue #819, opened 2021-04-05
Ambiguity in Instructions for running exercises on web template( For windows Users)


shashwat623:
Under Instructions for Web Templates on how to perform the exercise it mentions - "_Inside assets\websocket_address.js_" , change the variable _websocket_address_ to the IP address through which the container is connected", which roughly means the IP address of the container. This works fine for Linux and Mac users, but windows users run the container on top of a WSL(Windows subsystem for Linux), for them this _websocket_address_ would correspond to the IP address of the vEthernet adapter(WSL) and not the container IP address, as the container runs on top of WSL. 
A lot of newbies start open source on windows platform and this might be a little confusing for them if documentation is not provided.
Thank you.

shashwat623:
Thank you @dvalladaresv

PR #855, opened 2021-04-12
Updated launching procedure for Windows and a minor unnecessity in follow_line exercise

shashwat623:
1. Made changes to the if statements in Exercise.html as using OR( || ) operation with the same operands in both the cases seems unnecessary and ambiguous.

2. **Updated the launching procedure for Windows** - The documentation in the web site just mentions the websocket_address to be the IP Address through which the container is connected. This is straightforward for Linux, but in the case of Windows, the exercise runs on top of Windows Subsystem for Linux(WSL). This means the IP Address through which the container is connected is that of the vEthernet adapter(WSL).

### DIFF
--- a/exercises/color_filter/web-template/assets/console.js
+++ b/exercises/color_filter/web-template/assets/console.js
@@ -68,6 +68,6 @@ python_console.addEventListener("keyup", function(event){
 
 // Execute a function when clicked
 python_console.addEventListener("click", function(event){
-	// Focus on the input that should current be active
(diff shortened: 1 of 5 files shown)


shashwat623:
Can anyone update me on this?

shashwat623:
@dvalladaresv Please update me on this.

dvalladaresv:
Thank you for your contribution @shashwat623 
Fix #819

PR #856, opened 2021-04-14
Fix-#806(modified launch file)

RUFFY-369:
issue #806 
changed the launch file to use drone_assets package's turtlebot3 xacro rather than from copied one


pariaspe (changes requested):
Launch fails, xacro file is not found. Please review gazebo environment variables.

fail-856

RUFFY-369:
@pariaspe  for drone_assets package, I don't think that if urdf is as an installation directory?

pariaspe:
> @pariaspe for drone_assets package, I don't think that if urdf is as an installation directory?

I already solved in JdeRobot/drones#78

RUFFY-369:
@pariaspe , yeah saw it, should I close this PR then as you already did it in #857 ?

pariaspe:
Yes, I also close this PR. Thanks for the help :)

PR #874, opened 2021-04-19
Duplicated user function definitions

ReyDoran:
User functions defined inside sequential part don't have the right scope, they can't access global variables, imports nor other functions (issue-870). This does not happen on functions defined inside the iterative part.
The solution I propose is to duplicate the user function definitions (modifying the code sent) from the sequential part to the iterative part. This partially solves the problem:
- functions defined on sequential part and called on iterative part will work correctly
- functions defined on sequential part and called on sequential part will still have the same scope problems

### DIFF
--- a/exercises/follow_line/web-template/exercise.py
+++ b/exercises/follow_line/web-template/exercise.py
@@ -119,15 +119,6 @@ class Template:
             iterative_code = re.sub(r'[^ ]while\(True\):|[^ ]while True:', '', iterative_code)
             iterative_code = re.sub(r'^[ ]{4}', '', iterative_code, flags=re.M)
 
-            # Duplicate function definitions from the sequential part to the iterative part (solves scope errors)
-            funcHeaderMatches = re.finditer(r'def.*:\n', sequential_code)  # Find all headers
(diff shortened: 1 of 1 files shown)


SakshayMahna:
This problem arises due to global and local dictionary passed to the `exec` function that we use for running the student code in background. Your solution seems promising, but there exists a much simpler fix.

This problem can be fixed in the file `exercise.py`:

- In the `process_code` function, remove the expression `reference_environment = {}`.
- Add a new line after line 148, and add the definition: `reference_environment = {"GUI": gui_module, "HAL": hal_module}`
- Replace, this expression: `exec(sequential_code, {"GUI": gui_module, "HAL": hal_module, "time": time}, reference_environment)` with this: `exec(sequential_code, reference_environment)`

Passing an empty dictionary as the second parameter to `exec` was causing this problem. The above steps should fix this. Do test it and let me know if it works.

ReyDoran:
Yes, I tested it and those changes fix all the scope problems. Thanks for the information. This branch now includes this fix instead of the previous one.

SakshayMahna (approved):

SakshayMahna:
Cool!

Issue #876, opened 2021-04-19
Loading and Saving code from the web template


shashwat623:
I was recently going through the code base of the exercises running on the web template(color_filter). I found something strange, while loading  and saving the code from the in browser editor we are unnecessarily using the `Exercise.py` application running on the server/container as an intermediary and thus increasing the number of websocket calls. 

To save the code written on the in-template editor, we are sending the code to `Exercise.py`  **from** ` ws_code.js`  and then saving it on the host machine. 
 
// saveCode() function inside `ws_code.js`
```
         function saveCode(){
	// Get the code from editor and add header

	var python_code = editor.getValue();
	python_code = "#save" + python_code;
	console.log("Code Sent! Check terminal for more information!");
	websocket_code.send(python_code)   // websocket send call to exercise.py
        }

```
Similarly to load the code from host machine we are sending a load signal to `Exercise.py`  **from**  `ws_code.js`.  `Exercise.py` is then loading the code from host machine and **sending it back to**  `ws_code.js`

// loadCode function inside `ws_code.js`

```
function loadCode(){
	// Send message to initiate load message
	var message = "#load";
	websocket_code.send(message);  // sending load code signal to exercise.py
}
```

` ws_code.js`  is running on the client side itself and doesn't need `Exercise.py` running on server/container to handle load/save code requests for it. Aren't we just unnecessarily using extra websocket calls because of this? 
Should I go ahead and place a pull request with the required changes in place? Please have a look at this and let me know @dvalladaresv @jmplaza @SakshayMahna @pmvera 

Thank you!!

SakshayMahna:
Hello!

Nice observation. But, I think we are not using these currently to save and load the code. Since, we are saving and loading from the Docker Image, these files would eventually get deleted if the a running container is removed. In place of that, we are using Javascript functions to save/load them on/from our local computer. This way the files would remain persistent.

Probably, these functions were not deleted when we shifted to the other mechanism.

shashwat623:
Thanks for your reply Sakshay.
Yes, I believe you are correct as the `saveCode()` and `loadCode()` functions inside `ws_code.js` file **are no longer needed**, in fact after analyzing the code base of the application, I have come to a conclusion that these functions are never even even being called. They must have been used somewhere in the previous architectures and are of no use now. To load/save code to/from the web template the `local_functions.js` file is instead used as of now, which implements event listeners for any click on the load or save button on the web template. 
[...]

PR #846, opened 2021-04-10
Follow line exercise for Noetic and Gazebo-11

Suhas-G:
This PR fixes #814 
It has modifications for ROS noetic, Gazebo-11 and Python 3.
This Dockerfile still user `melodic-devel` branch of CustomRobots since the `noetic-devel` branch is giving some issue.

PR #879, opened 2021-04-19
Removed saveCode() and loadCode() functions for exercises running via the web template

shashwat623:
This pull request is in reference to #876 

Removed `saveCode()` and `loadCode()` functions as they are no longer needed for execution from web template. This is because these functions hand over the save/load procedure to `exercise.py`, which is running on the docker container. Saving and loading from the Docker Image wouldn't make sense as those files would be deleted when a running container is removed.
The `local_functions.js` file is instead used now to save/load code to/from the web template.

### DIFF
--- a/exercises/color_filter/web-template/assets/ws_code.js
+++ b/exercises/color_filter/web-template/assets/ws_code.js
@@ -81,24 +81,6 @@ function stopCode(){
 	running = false;
 }
 
-// Function to save the code
-function saveCode(){
-	// Get the code from editor and add header
-
-	var python_code = editor.getValue();
-	python_code = "#save" + python_code;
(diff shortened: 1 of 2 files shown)


dvalladaresv:
Very well observed and thank you very much for the contribution! @shashwat623 
Fix #879

shashwat623:
Thank you for your reviews @dvalladaresv @jmplaza 
Looking forward to contributing more.

jmplaza:
HI @shashwat623 , betatesting [pull-request #846](https://github.com/JdeRobot/RoboticsAcademy/pull/846) would be a nice contribution

shashwat623:
Alright @jmplaza, I'm on it.

PR #896, opened 2021-04-28
Added port number 1108 for console connection to web template.

shashwat623:
Added port number 1108 for console connection to web template. Without it the web template console will not connect in few exercises like line-follow.


pariaspe:
Already included, thanks anyway

Issue #903, opened 2021-04-29
[Performance] Gzclient Gazebo 3D viewer started (and finished) on demand


jmplaza:
and so no computation will be wasted on it when the user decides not to start that widget in the exercise webpage

ReyDoran:
Hi, I'm going to extend on demand gzclient to noetic branch

PR #899, opened 2021-04-29
Django webserver inside RADI - Issue 875

ReyDoran:
Includes a Django server which serves the exercise pages. It will allow refactoring web templates to share common parts and making easier to update the frontend ([issue-813](https://github.com/JdeRobot/RoboticsAcademy/issues/813)). The main change in the folders structure is that the previous exercises/ content is now within exercises/static/exercises/, as Django is serving it. Also, RoboticsAcademy.html is now a Django template and is located in exercises/templates/exercises/.
Exercises are now added by placing the folder exercise_name inside exercises/static/exercises and inlcuding it in db.sqlite3. This can be done by using the Django admin site:
1) Start the server
python3.8 manage.py runserver
2) Log in to the admin site. In the browser go to 127.0.0.1/admin, account “user”, password “pass”.
3) Click 'new exercise' and fill: exercise id (folder name), name (name to display), state (active), language, and description (description to display). Save and exit.
4) Commit db.sqlite3 changes to the repository.

The RADI can be launched now with this command:
docker run -it -p 8000:8000 -p 8080:8080 -p 7681:7681 -p 2303:2303 -p 1905:1905 -p 8765:8765 -p 6080:6080 -p 1108:1108 jderobot/robotics-academy ./start.sh
and it will run both the Django webserver and the Robotics Academy backend (manager.py). The RoboticsAcademy.html will be available in the browser at 127.0.0.0:8000.
If the RADI is launched with the old command:
docker run -it -p 8080:8080 -p 7681:7681 -p 2303:2303 -p 1905:1905 -p 8765:8765 -p 6080:6080 -p 1108:1108 jderobot/robotics-academy  python3.8 manager.py
it will only run the backend.

PR #919, opened 2021-05-03
on-demand gz client and console (refs #903)

Suhas-G:
This PR is a draft fix for #903 
If this approach is fine, I will clean up the code a bit and will do the same for other exercises.

Apart from that, I noticed that current [manager.py](https://github.com/JdeRobot/RoboticsAcademy/blob/51d21c2107bc4ee183e7a7fb94033aa0d337646a/scripts/manager.py#L20) have file paths without static.

This commit https://github.com/JdeRobot/RoboticsAcademy/pull/899/commits/b042f4fb7bcd447cec7d73e6d0f2687ea431b50c
had introduced different paths for instructions, since the folder structure has changed. I see that the folder structure is still same in master branch but the instructions have been reverted. @SakshayMahna / @jmplaza Can you please review?

### DIFF
--- a/exercises/static/exercises/follow_line/web-template/assets/launcher.js
+++ b/exercises/static/exercises/follow_line/web-template/assets/launcher.js
@@ -1,3 +1,6 @@
+var ws_manager, exercise;
+var callbacks = {};
+
 function startSim() {
     ws_manager = new WebSocket("ws://" + websocket_address + ":8765/");
     exercise = "follow_line"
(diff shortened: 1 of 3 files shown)


SakshayMahna:
Hello!
I went through your PR. It looks interesting.

With this Issue I was also thinking of refactoring the `manager.py` file. It has experienced the most change and in a rough manner. Probably, we can have a proper class based implementation like we have for exercise.py.

Increasing usage of manager.py (like starting and closing gzclient on demand) would require the websocket to be kept alive for the whole duration. This will in turn require acknowledgement messages(or ping pong messages)(I think I saw them in the draft changes as well), and this would make the code really cluttered. That's why for the refactoring.

Maybe @jmplaza could comment better on it.

However, current challenge remains to integrate GPU usage with the container, which is more important currently.

SakshayMahna:
New issue for the refactor. #950

Issue #885, opened 2021-04-21
[drones] Update drone exercises to RADI v2.3.18


iamarkaj:
- The command console.print() doesn’t work.
- The ‘Stop’ and ‘Reset’ buttons don’t work properly either.

Original post can be found [here](https://forum.jderobot.org/t/feedback-about-drone-cat-and-mouse-exercise/158)

- Delete console messages after every resubmission of student's code
- Add and modify developer launching instructions for drone-cat-mouse, follow-turtlebot, follow-road exercises

pariaspe:
Real Time Factor (RTF) should be added too. 

Updated exercises should look like:
![last_version](https://user-images.githubusercontent.com/35956525/117568327-f3c5ca00-b0bf-11eb-8327-afc2365161c9.png)

pariaspe:
## Drone Cat Mouse
- [ ] The command console.print() doesn’t work.
- [ ] The ‘Stop’ and ‘Reset’ buttons don’t work properly either.
- [ ] Delete console messages after every re-submission of student's code
- [ ] Add RTF

![catmouse](https://user-images.githubusercontent.com/35956525/117568433-78b0e380-b0c0-11eb-9e4a-773eb400bdb1.png)

## Follow Turtlebot
- [ ] The command console.print() doesn’t work.
- [ ] Delete console messages after every re-submission of student's code
- [ ] Add RTF

![follow_ttb](https://user-images.githubusercontent.com/35956525/117568437-7ea6c480-b0c0-11eb-8bce-8f379c669300.png)

## Follow Road
- [ ] Code editor seems broken.
- [ ] Connection isn't established.
- [ ] The command console.print() doesn’t work.
- [ ] Delete console messages after every re-submission of student's code
- [ ] Add RTF

![follow_road](https://user-images.githubusercontent.com/35956525/117568474-b6157100-b0c0-11eb-83c2-3633837675f7.png)

pariaspe:
Fixed in #944. Follow_road world might be revisited since gas_station model has been temporally disable cause an error while gazebo loading

PR #927, opened 2021-05-05
[drones] Updated the drone exercises

iamarkaj:
fixes issue #885


iamarkaj:
Currently, the drone crashes sometimes upon pressing the Reset button. Although, pressing the button again fixes this and resets the world. But this needs to be fixed.

@pariaspe pinging for review

PR #928, opened 2021-05-05
Add new DL Digit Classifier exercise

### LINKED ISSUE #915: [DeepLearning] Initial prototype of an exercise about DeepLearning for image classification

dpascualhe:
Closes #915


dpascualhe:
This PR adds the new `digit classification using deep learning` exercise to the Robotics Academy base. Its main difference with respect to previous exercises is that it replaces the code editor by a single button to upload a trained deep learning model, which is transferred from the client side and used for processing webcam images in real-time in the host. 

It only works for Python 3 due to [onnxruntime](https://www.onnxruntime.ai/index.html), which is used to load the model uploaded by the user and inference. It has huge interoperability, but lacks support for Python2 (that's why this PR's target branch is `noetic`). All requirements for this exercise, including onnxruntime, are installed using pip.

@pariaspe, @jmplaza told me that you're currently working in releasing the new RADI 3.0. If we can include the new dependencies so that the new exercise can work with the new RADI, that'd be awesome! Let me know if you have any concerns.

PR #944, opened 2021-05-10
Issue 885

iamarkaj:
- console.print() is not required anymore
- I suppose deletion of console messages is not necessary now. However, typing `clear` / `cls` inside the console will work.
- There seems to be some problem with the gas station model inside the follow-road exercise.

@pariaspe pinging for review


pariaspe:
No, console.log() is not required anymore. New versions can use print(). I agree, cleaning the console is not really important.

Yes, gazebo can't find locally gas station model. I'll try to review it tomorrow night, thanks.

pariaspe:
Hi @iamarkaj,

- RTF is not working properly. Label shows a zero value always.
- There is no need to change file paths since the RADI copy them to old location (check [this](https://github.com/JdeRobot/RoboticsAcademy/blob/2c17d2e3ee67d8b1f69612b8660b9a02830eeca5/scripts/Dockerfile#L232) out).
- `code/academy.py` doesn't seem an unnecessary file to me. We should keep them.
- I'm not sure if it is worth enough to maintain developer instructions. Robotics Academy is changing a lot and these instructions get old really fast, what do you think?

jmplaza:
Removing the "developer instructions" for launching the exercises makes RoboticsAcademy easier to maintain. The best way to launch them is just using manager.py, that will always be up to date. For instance currently gzweb is deprecated. In case of any developer interested in launching some exercises in a different way, she/he may check the steps that manager.py follows and try different things.

In fact we should also remove the instructions for ROS node release of the exercises for those already available in Web based release (>=2.3). The simpler the better

iamarkaj:
RTF changes as soon as the code is submitted. 
My bad I did not notice [this](https://github.com/JdeRobot/RoboticsAcademy/blob/2c17d2e3ee67d8b1f69612b8660b9a02830eeca5/scripts/Dockerfile#L232).

@pariaspe pinging for review

pariaspe:
Merging then, thanks @iamarkaj

Issue #1007, opened 2021-06-09
[GitHub Pages] Remove ROSnode instructions from migrated drone exercises


pariaspe:
Instructions previous to 2.3 version should be removed since those version are getting deprecated. 
Only applies to exercises already migrated.

pariaspe:
Drone-cat-mouse updated in #1009 
Follow_turtlebot and follow_road to go. Both need RTF display to be added.

pariaspe:
Solved in #1009 and #1022

PR #1022, opened 2021-06-12
Fixed RTF and updated docs of drone exercises

iamarkaj:
Closes issue #1007


iamarkaj:
In case RTF does not work for follow_turtlebot or follow_road, clear website cache

pariaspe (approved):

PR #1037, opened 2021-06-18
Source installation of jderobot drones inside RADI

iamarkaj:
closes issue #1030


pariaspe (approved):
Without sourcing drone_assets path is pointing to binaries location (/opt/ros/melodic...) instead of catkin location

Issue #1056, opened 2021-06-25
[RADI-2.4] Add web-template of Rescue People exercise


iamarkaj:

pariaspe:
Migration to 2.4 aborted. Exercises migrated directly to 3.1 (see #1070)

PR #1057, opened 2021-06-25
added web-template of rescue people exercise

iamarkaj:
Closes issue #1056


pariaspe (approved):
Exercise runs as expected. However, gazebo VNC window opens and remains black for 20-30 seconds. It doesn't seems to be related with model loading.

PR #1078, opened 2021-07-01
Updated docker image build instructions for noetic branch

shashwat623:
There's no deafult `Dockerfile` in the scripts folder. To reflect to `Dockerfile-noetic-3.1` we need to run `build.sh`, otherwise it shows an error.

Screenshot from 2021-07-01 11-03-27

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -25,8 +25,8 @@ First you need to build the image. Then, you need to run a container.
 ```
 git clone https://github.com/JdeRobot/RoboticsAcademy.git -b noetic
 cd scripts
-docker build -t image-name .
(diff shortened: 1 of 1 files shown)


pariaspe:
Tag is needed. Naming version protocol is the same to melodic versions. We are currently in 3.1.2.

shashwat623:
@pariaspe  Yes, tag is needed. My issue is with using  `docker build -t image-name .` instead of `./build.sh <tag>`. The former command gives an error as there is no default `Dockerfile`. Moreover, the build command inside `./build.sh` defaults the tag to 3.1 so adding tag after it wont really matter. The default tag would however be required in the `docker run` command

pariaspe:
You can select the Dockerfile to build with `-f` flag.
```
docker build -t jderobot/robotics-academy:3.1.2 --file Dockerfile-noetic-3.1 .
```
Tag does matter when pushing the image to the dockerhub, so it has to be included.

shashwat623:
Alright @pariaspe Thanks a ton !!

PR #1105, opened 2021-07-06
Extend on demand gzclient to other exercises

SakshayMahna:


SakshayMahna:
In reference to issue #903 

On Demand Gzclient extended to other exercises as well.

Issue #1096, opened 2021-07-03
[RADI] Connection closed automatically after a while of the stablishment


pariaspe:
Connection get closed due to a Timeout Error. This screenshot was taken using RADI v3.1.2

![timeout](https://user-images.githubusercontent.com/35956525/124353556-7f169400-dc07-11eb-844c-3a444159b91d.png)

jmplaza:
It is weird to see Python-2.7 in SocketServer, maybe we should upgrade to its Python-3.X release. 

RADI  3.1 works with Ubuntu 20.04 as underlying linux distribution (FROM nvidia/opengl:1.2-glvnd-runtime-ubuntu20.04) and ROS Noetic, which does work with Python 3.X.  There is no need to use Python-2.7 for anything.

In addtion, in the RADI-3.1.2 docker file I see
# websocket server dependency
RUN pip3 install websocket_server posix-ipc django
RUN python3.8 -m pip install websockets asyncio

but from the snapshot you attached it seems it is using websocket server with Python-2.7...

iamarkaj:
![timeout-error](https://user-images.githubusercontent.com/40225765/124371029-812a3280-dc9b-11eb-977d-d29d08751bc3.png)

I get the same timeout error after 30 seconds but the python version is different for me in RADI-3.1.2

PR #1110, opened 2021-07-06
[noetic] Fixes connection timeout

iamarkaj:
Draft PR to close issue #1096


pariaspe:
"Connection closed" alert does not appear anymore since it has been commented out. However, timeout exception keeps showing up. I have not observed that the error affects the execution of the exercise, but we should catch and process it.

aa-timeout

pariaspe (approved):

Issue #1121, opened 2021-07-08
[drone-cat-mouse] Add automatic evaluators


pariaspe:

pariaspe:
First version of evaluator merged. Closing..

PR #1124, opened 2021-07-09
Automatic evaluator for drone-cat-mouse

iamarkaj:
Draft PR to close issue #1121


pariaspe:
It looks good to me, good job @iamarkaj ! Some questions about it: Which is the score policy? Does the evaluation start after the "Evaluate" btn is pressed, isn't it? When does it finish?
@jmplaza any further suggestion?

evaluator

iamarkaj:
Hi @pariaspe 

**Scoring policy**:
*dist = distance between the position of cat and drone*

1. if dist is less than 2 , raise score by 0.5
2. if dist is more than 2 and less than 3, raise score by 0.25
3. if dist is more than 3 and less than 4, raise score by 0.12
4. else raise score by 0.01

Notes: 
1. I have added relevant comments now, you can recheck the calc_score function inside exercise.py
2. The plotting starts by pressing the **Evaluate** button and stops by pressing the same button again
3. If you hover over the graph, you will get access to few more controls like zoom in, zoom out, auto-rescale and save

jmplaza:
I would the graphics with more natural axis: X+ = time, Y+= instantaneous distance showing only positive values (Y=dist, and then Y=0 means dist=0).
For the sake fare evaluation I propose a "measuring interval" (let's say 2 minutes, instead of waiting till the second click on the evaluate button) and "measuring points" (maybe one each 500ms). And so, at each measuring point the score is updated depending on the dist. The general criteria is the higher increment the lower the dist:

1. if dist is less than 2 , raise score by 0.5
2. if dist is more than 2 and less than 3, raise score by 0.25
3. if dist is more than 3 and less than 4, raise score by 0.12
4. else not raise score

The measuring interval may be shorter if the user clicks a second time at evaluate time, but its maximum is 120 seconds. This way we may compare performance of several RobotisAcademy users in a clean way, and even show the monthly record or "world record" for this exercise :-)

There would be 2minutes = 120 secs, 120secs = 240 measuring points. We may use another increments in order to bound the maximum score to 100 points... We could go with this prototype and then check it working with more users. We may further adapt the dist thresholds or the frequency of the measuring points.

What do you think?

pariaspe:
It's looking good! Some questions and issues to fix:
- I don't understand Y axis. Right after connection establishment distance is close to 0 (?) and if I move one drone closer to the other the distance increase (?).
- Score label remains always as zero value.
- I don't understand Evaluate button either. First click seems logic to me, but afterwards I don't understand why time interval is reduced. I would suggest just to reset the evaluation (score and graph).
- I've also observed that distances also need some adjustments too. In the image drones are almost touching and the distance is 1.5 (meters I guess and if I'm understanding Y axis correctly). I suppose that distance is estimated from drone centroids, then drone radius should be subtracted before calculating the score.
- Does it need the Play button to be pressed to calculate the score? After bringing drones closer, I've just seen some updates on the graph after clicking on Play button.

a

iamarkaj:
**Evaluate button:**
- When pressed for the first time it shows a graph plotting the distance between the cat and mouse drone
dist=\sqrt{{(x_{mouse}-x_{cat})}^2+{(y_{mouse}-y_{cat})}^2+{(z_{mouse}-z_{cat})}^2}^2+{(y_{mouse}-y_{cat})}^2+{(z_{mouse}-z_{cat})}^2)
- **x-axis:** Time in seconds **y-axis:** Distance in metres
- Not necessary to submit the code. In that case both the drones will be on the ground and it will plot a horizontal line indicating the fixed distance between them
- Measuring interval is 2 minutes. On subsequent pressing the button it reduces the measuring interval by 15 seconds each time up to a minimum measuring interval of 30 seconds
- The measuring interval resets each time the user's code is submitted. Press the **Play** and **Evaluate** button sequentially
- In case to stop the plotting press **Stop** button
- The background color of the plot is green and changes to red when dist is greater than 4

*Note: This is experimental and can be changed if required*


**Score:**
- The scoring starts as soon as both the drones are up in the air
**Condition:** both the z values of cat and drone must be greater than 0.25
**Reason:** For fair evaluation
**Implies:** To bring the cat drone in the air, you have to submit the code
- **Scoring policy:**
  1. if dist is less than 2 , raise score by 0.5
  2. if dist is more than 2 and less than 3, raise score by 0.25
  3. if dist is more than 3 and less than 4, raise score by 0.12
  4. else not raise score
- **Maximum Score:** 100
- **Measuring interval:** 2 minutes
- The score resets each time the user's code is submitted

pariaspe:
Some notes about:
- I'd suggest to do some changes with the Evaluation button. I'd change it to a two-states button. The button could be pressed or not. Being pressed would imply that an evaluation is on progress, otherwise the evaluation is not done (start evaluation or stop evaluation). This implies some changes: (1) Time interval will be fixed to 120s but it can be stopped before by hitting the button again. (2) Graph could be always visible or stay hidden until the first click and keep visible after it. (3) Clicking evaluate button could imply also a 'Play Code' click, and start executing the code if it's not already running.
- All other changes on the evaluate button and graph looks cool to me.
- The score seems nice to me too. I'd only suggest that the score also reset itself when the Evaluate button is pressed (but not when is unpressed, allowing the user to save it).

iamarkaj:
Closing this. Opened a new PR #1202

PR #1137, opened 2021-07-14
Stop and resume functions controlled by manager

ReyDoran:
manager.py controlls stop and resume functions. I tried to use similar functions. I only changed `running` variable set from controller.js to launcher.js, as it could create an error (running variable set to false when running, or to true when stopped) if the buttons were clicked two times before the "Ping" message was sent.

### DIFF
--- a/exercises/static/exercises/follow_line/web-template/assets/controller.js
+++ b/exercises/static/exercises/follow_line/web-template/assets/controller.js
@@ -17,6 +17,9 @@ function start(){
 
 // Function to stop the student solution
 function stop(){
+    // Code Websocket
+    stopCode();
+    
     // Manager Websocket
     if (running == true) {
         stopSimulation();
@@ -28,6 +31,9 @@ function stop(){
 
 // Function to reset the simulation
 function reset(){
+    // Code Websocket
+    stopCode();
+
     // Manager Websocket
     resetSimulation();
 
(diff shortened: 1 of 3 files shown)


exercises/static/exercises/follow_line/web-template/assets/controller.js
```diff
@@ -1,35 +1,33 @@
...
     unpause_lap();
-
-    // Mark the variables
-    running = true;
 }
 
 // Function to stop the student solution
 function stop(){
-    // Code Websocket
-    stopCode();
```
SakshayMahna, line 19: In this part, shouldn't we keep the `stopCode()` function?

If we remove it, then on clicking Stop, the simulation will pause, however the student code will keep on running.

exercises/static/exercises/follow_line/web-template/assets/controller.js
```diff
@@ -1,35 +1,33 @@
...
     pause_lap();
-
-    // Mark the variables
-    running = false;
 }
 
 // Function to reset the simulation
 function reset(){
-    // Code websocket
-    stopCode();
```
SakshayMahna, line 31: Not sure about this, but does the reset work fine after removing this?

There are 2 test cases for reset:
- Pause and then reset
- Reset a running code(the car will reset it's position and start running the same code)

Does the Lap Time and everything else work fine without this?

SakshayMahna:
Other than that, everything looks good to go!

ReyDoran:
The test cases work, but it is true that the code keeps running. I'm going to restore stopCode() call and merge it to issue-1089 and to master afterwards. Thanks!

PR #1163, opened 2021-07-17
Updated the docs of pick and place exercise

predator4hack:


sixer51:
It seems that Gem for windows system is also added, but it does not influence any other thing, so I would agree to merge it.

PR #1183, opened 2021-07-21
Added Teleoperator mode to follow_line exercise

dhrodao:
Added teleoperator mode to follow_line exercise


ReyDoran:
Hi, I just tested this branch. Messages seem to be sent (console displays v and w) but the car does not move when using the keyboard arrows.

dhrodao:
I tested it yesterday and it was working. Did you pause the scenario before clicking the teleoperator button?

ReyDoran:
Yes, I tried pausing. The steps I follow in order to test it is: build the image downloading the branch issue-1177 of RoboticsAcademy instead of master, run the image with ./start.sh and load the exercise. I send some code, pause and press teleoperate.

dhrodao:
The problem is pausing, because gazebo physics are also getting paused, so, just press start and then press the teleoperator button, then, you will see that the car stops and you will be able to use the arrow keys.

Issue #1204, opened 2021-08-02
[Machine Vision] Migration of machine vision exercise to web-templates


predator4hack:

jmplaza:
There are already two exercises in the new web-templates: ColorFilter and OpticalFlow teleop

PR #1207, opened 2021-08-02
Machine vision

predator4hack:
solves #1204


predator4hack:
Started migration of machine vision exercise, contains ace-builds files

PR #1216, opened 2021-08-08
Fix GUI and HAL modules generation

dhrodao:
follow_line_game is now part of the circuit exercises.
before checking code, now its checked if the exercise is part of the follow_line challenge.


dhrodao:
follow_line_game is now part of the circuit exercises.
before checking code, now its checked if the exercise is part of the follow_line challenge.

dhrodao:
follow_line_game is now part of the circuit exercises.
before checking code, now its checked if the exercise is part of the follow_line challenge.

Issue #1345, opened 2021-10-21
Update log storing method


pariaspe:
Current method appends roslaunchs output to custom file.
Log storing can be done directly using `--log` option in `roslaunch` command. One log will be generated for each roslaunch. Logs location can be found running `roslaunch-logs`.

If parsing output is needed, `roslaunch-logs` can be used to locate where logs are stored (`roscore` must be running) and read the searched file (#1335).

**Source:**
[roslaunch](http://wiki.ros.org/roslaunch/Commandline%20Tools)

pariaspe:
Implemented in drone exercises (#1348). Documentation must be also updated.
It might be extended to the whole set of exercises.

pariaspe:
Old ROS log generated files should be removed when launching an exercises to avoid filling up hard-disk on long-live containers,
Command `rosclean purge -y` might do the job.

PR #1369, opened 2021-11-06
Clear existing ROS logs

iamarkaj:
Closes issue #1345


pariaspe:
`start_logs.sh` is only executed when starting the docker container, isn't it? Then logs are only removed once, instead of removing them before each exercise launching.

Issue #1368, opened 2021-11-06
[Drones] New Package Delivery exercise


iamarkaj:
- [ ] Exercise prototype
- [ ] Documentation

PR #1378, opened 2021-11-10
[Package Delivery] Added new drone exercise

iamarkaj:
Closes issue #1368

package_delivery_teaser

package_delivery


pariaspe:
Drone can't takeoff using web-templates version. It does takeoff using ROS-node version.

error

pariaspe:
ROS Time Moved Backwards exception is fixed in jdeRobot/drones#157

PR #1383, opened 2021-11-12
Redesigned Rescue People

pariaspe:
#1382 

rescue_p


pariaspe:
Tested and works fine

PR #1583, opened 2022-03-19
[ROS 2] improvement of Dockerfile

Carlosalpha1:
Hi,
I have improved Dockerfile-foxy.base and Dockerfile-foxy files


trunc8:
Hi @Carlosalpha1, you can be more detailed with the Pull Request description. For example, instead of "improved Dockerfile-foxy.base and Dockerfile-foxy files," you could mention why certain changes were made and why other lines of code were added. Here, for instance, you can describe in two parts here:
- URL/version number of packages have changed at the remote hosting site. This has been fixed in Dockerfile
- Added more ROS packages to be used by the new FollowPerson exercise

This will give the idea to future contributors as to why certain changes were necessary

PR #1686, opened 2022-04-26
Added --no-check-certificate to RADI base image

### LINKED ISSUE #1685: RADI Base Image vncSelf-signed certificate encountered.
while building the RADI's Base Image sometimes Turbo vnc server causes the following error:-

`vncSelf-signed certificate encountered. `

akshaynarisetti:
Sometimes while building Dockerfile.base the vnc server `vncSelf-signed certificate error` is encountered.

To bypass this error "--no-check-certificate" flag has been added to the commands in RADI base Image.

It makes wget continue despite verification failure, but it still will complain - without that option set wget would simply stop once verification fails, causing the Docker Image build process to fail.

**This is added as a failsafe**

Fixes #1685


akshaynarisetti:
I have built and verified the Image, everything works as expected :)

PR #1767, opened 2022-07-08
centered an image in mobile view

sumanparajuli:
Google summer of code image was left aligned when viewing from a mobile device. I have centered it to look better.


pariaspe:
Hi @sumanparajuli, thanks for the contribution!

PR #1771, opened 2022-07-11
remove redundant and old files

Apoorvgarg-creator:
Removing assets folder present inside the web templates folder for every exercise


dvalladaresv:
Thanks for the contribution :)

PR #1795, opened 2022-08-09
[drones] multiprocessing prototypes

RUFFY-369:
to be used with pr #1772

PR #1867, opened 2022-11-13
Labyrinth Escape frontend

pariaspe:
From Prakarsh gsoc (#1795)


iamarkaj (changes requested):
**World before including this changes**

world_before_including_the_changes

**World after including this changes**

world_after_including_the_changes


According to me, the grass_plane looks better than asphalt_plane. Thus, we can close this PR.

Issue #1520, opened 2022-01-23
Unable to run RADI with GPU acceleration


pariaspe:
I'm getting this error with VGL when I'm trying to run the latest RADI with GPU acceleration:
```
[VGL] ERROR: Could not open display /dev/dri/card0.
[WARN] [1642851485.742479]: [GAZEBO] Check timeout exceeded
```

Since VGL fails, Gazebo launching also fails and I can not play with the exercise.
I was able to use GPU acceleration with RADI 3.1 versions.

soumilbaldota:
You could try to do this and see if it works.
```sudo nvidia-xconfig -a --allow-empty-initial-configuration --use-display-device=None \ --virtual=1920x1200 --busid {busid}```

soumilbaldota:
Similar to 
[link](https://github.com/VirtualGL/virtualgl/issues/108#issuecomment-531851472)

soumilbaldota:
Similar to (https://github.com/VirtualGL/virtualgl/issues/108#issuecomment-531851472)

pariaspe:
Hi @soumilbaldota, thanks for the suggestion but it didn't work in my system. Have you tried it?

pariaspe:
Solved in #1902. Thanks @iamarkaj !

PR #1902, opened 2023-01-08
Fix vgl error: Could not open display /dev/dri/card0

iamarkaj:
Solves issue #1520 

Check out the **third** [significant change relative to VirtualGL 2.6.5](https://github.com/VirtualGL/virtualgl/releases/tag/3.0beta1)

`
The EGL back end can be activated by setting the VGL_DISPLAY environment variable to the path of a DRI device, such as /dev/dri/card0, or by passing that device path to vglrun using the -d argument.
`


iamarkaj:
hi @pariaspe @dpascualhe, pinging for review

pariaspe (approved):
Tested and working.
Screenshot from 2023-01-09 14-58-15

I've observed a permission error after the "Starting Gzclient message" that we should also check. It happens also without using VGL.

Issue #1948, opened 2023-02-26
Optimization of Gazebo execution inside the RoboticsAcademy Docker Image (RADI) for Linux (+Nvidia graphics card)


jmplaza:

jmplaza:
+Nvidia graphics cards, or even Intel graphics cards

PR #1950, opened 2023-02-27
Make DRI device configurable to enable GPU acceleration 

dpascualhe:
As of now, DRI device address is hardcoded (`/dev/dri/card0`), and so users can't choose which GPU to use for acceleration.

In this PR, we include changes to allow users to set the environment variable `DRI_NAME` when running the Docker container. If a user has an integrated Intel GPU and a dedicated NVIDIA one, the user can set this new env variable accordingly and choose which GPU to use. Further instructions are provided in [gpu_acceleration.md](https://github.com/JdeRobot/RoboticsAcademy/blob/b24f67208427f05787c5db3dc7f6ee257e3b25f1/docs/gpu_acceleration.md).

Solves issue #1948

### DIFF
--- a/docs/gpu_acceleration.md
+++ b/docs/gpu_acceleration.md
@@ -1,0 +1,41 @@
+### [Back to main README.][]
+
+[Back to main README.]: ../README.md
+
+# GPU Acceleration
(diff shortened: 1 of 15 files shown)


pariaspe (approved):
Both intel and nvidia gpu acceleration is working in my machine. Exercise launching instructions should be updated.

PR #1983, opened 2023-03-23
updated docker ports

pawanw17:
- Removed 8080 here because point 5) runs django webserver which already opens 8080, trying to open it again in docker would cause this error :
image

- Added port 7163 for new RAM

@jmplaza @pariaspe


pariaspe (approved):
Good catch!

Issue #2027, opened 2023-04-29
Broken forum hyperlink on homepage


pawanw17:
![image](https://user-images.githubusercontent.com/40732709/235326577-55fcde60-2e68-4d09-9367-ee85a02fbbf7.png)
The button redirects to 
https://forum.jderobot.org/c/english/roboticsacademy/12
which is deprecated, instead it should redirect to
https://forum.unibotics.org/

Bhardwaj-Himanshu:
Hi @pawanw17,
This issue was an easy fix, just located the forum links and changed it to the provided forum link.

But I am stuck at `**Github/Git adding multiple files to the pull request, when I only want to add one.**`

To resolve this issue `I created a new branch, added the changes to it, commited it, and pushed it to my fork repo as another branch`, but it is still showing me other commits in other branches as addition.
No idea on what to do?
Just a small guess of doing git reset --hard SHA<id>, but if anything else that could solve this?

It looks like this-
![image](https://user-images.githubusercontent.com/74407676/235351048-5d2ea0a5-3c93-46ee-aab3-ecb403536e32.png)
![image](https://user-images.githubusercontent.com/74407676/235351051-c65008cc-9e9a-4397-b3c5-c1e31d7ea50d.png)

jmplaza:
Fixed in #2028

pawanw17:
Hi @jmplaza the issue still exists... I suspect the wrong file was changed...

jmplaza:
It has been changed at the repo, but no RADI has been released yet with it. So it will be solved in the next RADI release.

pawanw17:
I implemented his changes locally to test... I think the correct place should have been :    
https://github.com/JdeRobot/RoboticsAcademy/blob/3616a42b394989c9c2368fe4492bdc97a8d55834/react_frontend/src/components/DrawerAppBar.js#L28

jmplaza:
perfect, just go ahead with the PR and you will solve this minor issue.

pawanw17:
Hi @Bhardwaj-Himanshu Do you want to take this up?

Bhardwaj-Himanshu:
I would love to @pawanw17 , but you can contribute to it as well and let me know what are you upto!

pawanw17:
Hi @Bhardwaj-Himanshu Since you're already working on this, you can continue and finish it. I'd just suggest you test your changes before making the PR. All the best!

Bhardwaj-Himanshu:
[...]

PR #2028, opened 2023-04-30
#2027 fixed.

Bhardwaj-Himanshu:
This issue was leading to broken forum link as mentioned in #2027.

**Solution**
The solution was quite simple located the old href="https://forum.jderobot.org/c/english/roboticsacademy/12" and replaced it with new href="https://forum.unibotics.org/" which should fix it.

Let me know if you have any reviews or changes.


jmplaza:
Thanks @Bhardwaj-Himanshu !

Issue #2025, opened 2023-04-28
Some links redirecting to a 404 page


Bhardwaj-Himanshu:
**Description of Issue**
While clicking through the documentation guidelines as well as contribution guidelines, link [documentation guide lines](https://jderobot.github.io/RoboticsAcademy/contribute/) is broken inside `contributing.md`.

https://user-images.githubusercontent.com/74407676/235203274-fee4d14d-419d-41d4-9a2b-3cc1a52d4de9.mp4

**What impact this change could cause?**
Causing a state of confusion and less retention of new users looking forward to work on the website.

**Fix?**
Just changing the links in `contributing.md` to new [Contributor Guide](https://jderobot.github.io/RoboticsAcademy/developer_guide)

jmplaza:
Fixed in #2029

PR #2029, opened 2023-04-30
#2025 fixed.

Bhardwaj-Himanshu:
Some links were redirecting to a 404 page which could be confusing for a new user as mentioned in #2025 .

Easy fix was to change some links in contributing.md so that they land on a page and don't show 404 error.


jmplaza:
Thanks @Bhardwaj-Himanshu

PR #2031, opened 2023-05-01
Unneeded gazebo path command on gzclient launch

Bhardwaj-Himanshu:
As discussed in issue #1904, there was an error when starting any exercise or gzclient, which was thrown due to `calling the file location`.

This pull request fixes it.

Here is what **BASE AND RADI IMAGE LOOK LIKE AFTER MAKING THE CHANGES**

https://github.com/JdeRobot/RoboticsAcademy/issues/1904#issuecomment-1528927544


pariaspe:
@jmplaza this leaves "gazebo_path" entry from [instruction.json](https://github.com/JdeRobot/RoboticsAcademy/blob/master/scripts/instructions.json) (old manager) unused. Do you remember why we need them? Should we remove them?

PR #2136, opened 2023-06-16
Update ASGI configuration for Django application

zahidhussain998:
This title accurately reflects the change made in the code, which involves updating the ASGI configuration for a Django application. It provides a concise summary of the modification made in the commit.


ango1994:
Hi @zahidhussain998. Can you provide some information on why this change is needed? The RADI is launched via entrypoint.sh which calls manage.py, so I don't see the need to call django.setup(), as manage.py calls it itself.

PR #2421, opened 2024-02-19
[Amazon Warehouse] Added warehouses with the Ackermann logistic robot to the database

lu164:
Added warehouses with the Ackermann logistic robot (World3 and World4) to the database, and I have also had to adjust some parameters in the map.py file so that the robot is correctly displayed on the map


OscarMrZ:
It seems it's due to the big size of the world and my pc, it loaded after 5 minutes!

Issue #2404, opened 2024-02-08
[Thin Templates Humble] GUI Images do not get updated


pawanw17:
https://github.com/JdeRobot/RoboticsAcademy/assets/40732709/b8e22f2b-8781-4bcb-b466-e17269699865

PR #2426, opened 2024-02-23
[Thin/Superthin templates] Initial migration of rescue_people and follow_person to superthin templates

OscarMrZ:
Initial testing with the new manager and HAL/GUI with the humble versions.


pawanw17:
closes #2404

Issue #2465, opened 2024-03-28
 Functionality issues in modified develop_academy.sh script


vaishnavimulik25:
![error](https://github.com/JdeRobot/RoboticsAcademy/assets/126563312/c22c48d0-e1bd-41ad-9e3f-9e3dce99669e)
I have followed the latest documentation and used the latest script for installation. Since I was getting this issue I tried manually running commands in develop.sh script. I found error during npm install.

vaishnavimulik25:
[Modified develop_academy.sh](https://github.com/JdeRobot/RoboticsAcademy/pull/2466)

OscarMrZ:
Hi @vaishnavimulik25, thanks for your time! Please describe your issue in more detail, why are you using npm? RA uses yarn as the package manager, and should always be used (as it is in the script). Do you also have errors with yarn?

Regarding

vaishnavimulik25:
Yes I had errors with yarn previously but after using [updated develop_academy.sh](https://github.com/JdeRobot/RoboticsAcademy/commit/cfa60b532e5fbb7e6d4a0245dd711925ffc91b79) I didn't had error.

OscarMrZ:
Outdated due to big changes in main branch. Please feel free to reopen if you think we are missing something!

PR #2466, opened 2024-03-28
Modified develop_academy.sh

vaishnavimulik25:
In [2465](https://github.com/JdeRobot/RoboticsAcademy/issues/2465) I faced issue while running develop_academy.sh. Following are the changes that I made.
1. User Interaction: Added functionality for the user to select Docker actions (start or stop) based on their preference. This improves usability and flexibility, allowing users to manage Docker containers more efficiently.
2. Dependency Management: Utilized npm install --force to address potential dependency tree issues that could hinder project functionality.

Testing:
Tested the updated script locally to verify that user prompts for Docker actions work as expected.
Verified that dependencies are installed correctly using npm install --force and that the script executes without errors.

working

modified

### DIFF
--- a/scripts/develop_academy.sh
+++ b/scripts/develop_academy.sh
@@ -39,6 +39,11 @@ if ! command -v nvm &> /dev/null; then
   source $NVM_DIR/nvm.sh;
 fi
 
+# Prepare yarn 
+if ! command -v yarn --version &> /dev/null; then
+  npm install --global yarn
+fi
+
 # Prepare the frontend
 nvm install 16
 nvm use 16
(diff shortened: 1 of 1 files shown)


OscarMrZ:
Hi @vaishnavimulik25, thanks for your contrib! Why do you think this interaction is needed exactly? Please describe a usecase.

Currently, thanks to ending the commands with ;, when the user closes the script with Crtl+C, docker-compose down and the rm commands get executed, efectively cleaning the developing environment. The user may start again with the command with no problems.

vaishnavimulik25:
Hi @OscarMrZ ,As a user, I encountered a problem where I lacked control over starting, stopping, and restarting Docker. Upon reviewing the script, I realized that I had to run it each time to execute docker-compose up, even when the development environment was already set up. This redundancy posed an inconvenience.

OscarMrZ:
Hi @vaishnavimulik25. Why did u need to restart the container so much? The script is intended to prepare a development environment, and then you can edit the code without having to rerun the script or docker-compose up. I think a more interesting approach would be to check if the react frontend should be regenerated (https://github.com/JdeRobot/RoboticsAcademy/issues/2470) . While the script will check if the env if setup each time that it is run, it only install the necessary things the first time, so I don't really think a while loop to check for docker reboot is needed, although it is a valid suggestion! 

Let me know if u would be interested in working in the mentioned issue!

javizqh:
Merged in https://github.com/JdeRobot/RoboticsAcademy/pull/3004

Issue #1998, opened 2023-04-06
[Enhancement] Hyperlink the JdeRobot Logo to homepage


pawanw17:
Currently to go back to exercises home in the RADI we have to use the back button of the browser, It'd be good if we could instead use the JDE Robot logo to do so.

![image](https://user-images.githubusercontent.com/40732709/230314667-387a9a42-f628-49af-a06f-3f1f999e2d1a.png)

pariaspe:
Thanks @pawanw17

damodardatta:
Hi,

I haven't been able to go back to the main page from exercises by pressing the logo while using the latest docker image.  Can you check if this repeats?

pariaspe:
Hi @damodardatta , please try with tags 3.4.30 and 4.4.32 instead of latest

damodardatta:
@pariaspe I've tried with the tags you suggested. I'm still not able to go back.

pariaspe:
Ok, reopening then

OscarMrZ:
Solved at #2467

pawanw17:
Reopening this because of the superthin refactoring

CDAM2020:
Made the pr #2524 fixing this

jmplaza:
Solved in #2524

PR #2467, opened 2024-03-29
added hyperlink to home

damodardatta:
solves #1998 and tested with RADI 4.4.32


OscarMrZ:
Nice @damodardatta! Merged!

PR #2512, opened 2024-05-03
Nvidia GPU support in automatic script

### LINKED ISSUE #2508: Add Nvidia support in automatic script
There are only for CPU and Intel GPU support in docker compose config, and also the ports implemented are from older versions, it would be good to change to the ports that are using actually in the last version.

Syekeon:


javizqh:
This has already been solved in this [PR](https://github.com/JdeRobot/RoboticsAcademy/pull/2676.)

PR #2523, opened 2024-05-09
FIX-issue-2406

CDAM2020:


pawanw17:
Hi, the fix looks good, but this PR doesn't need a change in docker-compose.yaml can you remove that file from this PR

CDAM2020:
Yeah, I didn't erase that file from the commit, sorry! I'll modify the pr

PR #2536, opened 2024-05-22
Migrating localized vacuum cleaner to ros2 humble

javizqh:
Migrating localized vacuum cleaner to ros2 humble


javizqh:
Need to test if map updates

javizqh:
Both this excercise and basic vacuum cleaner not updating the map. #2537

javizqh:
Map not reseting

javizqh:
I rebased in order to solve this conflict, but after doing it and updating the current database it still shows as a conflict

javizqh:
Moved to #2538

PR #2538, opened 2024-05-27
Migrating localized vacuum cleaner to ros2 humble [Rebased]

javizqh:
Map not reseting


javizqh:
The updated db is part of  #2539

javizqh:
Now maps resets, but a blue dot still remains in the map

javizqh:
https://github.com/JdeRobot/RoboticsAcademy/assets/124056014/1625d4b3-9090-4bef-b8d4-027e4970a4c4

PR #2565, opened 2024-06-07
[Montecarlo laser loc] Update interfaces and remove useless files

### LINKED ISSUE #2564: [Montecarlo laser loc] Update interfaces to hal_interfaces

javizqh:
Removed interfaces and shared folders and use hal_interfaces.
Resolves #2564


exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/HAL.py
```diff
@@ -1,48 +1,67 @@
...
+        ]
+    )
 
-def getBumperData():
-    bumper.spin_bumper_node()
-    return bumper.getBumperData()
+    # Spin nodes so that subscription callbacks load topic data
+    executor = rclpy.executors.MultiThreadedExecutor()
+    executor.add_node(odometry_node)
+    executor.add_node(laser_node)
```
pawanw17, line 35: Hi @javizqh Why is bumper node not added in the executor?

javizqh: As the bumper is not as essential as the other nodes and I think it is not used a lot in the solution that is why I kept the spinOnce method

javizqh: But if you want I could add it to the executor

pawanw17: oh okay, that makes sense...

PR #2588, opened 2024-06-21
[Amazon Warehouse] Updated Hints section documentation

lu164:
Added hints to solve the problem with geometric constraints of the robot


jmplaza:
Good job @lu164 !

PR #2600, opened 2024-06-28
[Basic Vacuum Cleaner] Two class GUI

### LINKED ISSUE #2537: [Vacuum Cleaner] Map not updating
This happens for the basic one and the localized one.

https://github.com/JdeRobot/RoboticsAcademy/assets/124056014/321f9690-2e20-46ac-9a77-a8f413145734

javizqh:
Switching from a one class GUI to the old 2 class approach because of some problems with the rendering with the new method.

Fixes #2537


pawanw17:
Hi @javizqh, please remove the commented out old code, if we want to recover it in the future then we will revert this commit

javizqh:
I have removed it

PR #2602, opened 2024-06-28
Updated the dockerfiles for Gazebo11 installation along with Gz harmonic

Arcane-01:
For issue #2569


pariaspe (approved):
Both exercises can coexist with ros dependencies now. Still pending to test an existing exercise

PR #2637, opened 2024-07-08
[Loc Vacuum Cleaner] Adding back shared image

javizqh:
Loc Vacuum Cleaner: Adding back shared image to display image


dduro2020:
Fixed bug in getMap() function

javizqh:
Thanks to @dduro2020 for providing the getMap function

javizqh:
This has already been solved in #2640 so this is no longer needed

PR #2656, opened 2024-07-12
[Rescue People] Fix stuttering in images

### LINKED ISSUE #2652: [Rescue People] Remove ack messages sent from the widget
This makes the images shown in the UI laggy.

javizqh:
The issue was not caused by the ack but from the HAL frequency rate. Increasing it to 30 Hz solves this issue.

Resolves #2652


pawanw17:
I tested it, it is still a bit laggy, it used to be a lot more fluid

PR #2657, opened 2024-07-12
Remove launchers and entry_points from exercises

javizqh:
This launchers and entry points were no longer needed with the new RAM as they are now in RI


jmplaza:
The launchers are now at RoboticsInfrastructure repo, in order to be usable from other applications such as BT-Studio or VisualCircuit, not only RoboticsAcademy

PR #2674, opened 2024-07-29
Fixed alertMessage timer and added fade effect

### LINKED ISSUE #2673: Fix auto-delete for showAlert message
After the timeout, no more new messages will be displayed.

espilya:


javizqh (approved):
Works, but right now does not trigger for syntactic errors. To fix this issue is necessary this https://github.com/JdeRobot/RoboticsApplicationManager/pull/141

PR #2676, opened 2024-07-30
[Work in progress] Add Nvidia support in automatic script 

### LINKED ISSUE #2508: Add Nvidia support in automatic script

dduro2020:


javizqh (changes requested):
Change the docker image to 4.6.4

javizqh (approved):

PR #2683, opened 2024-07-31
[global-navigation] Improve map <-> world coordinates conversion

ghost:
Fixed some of the conversion functions in the map module and added two functions to convert between map (numpy grid) and world (gazebo) coordinates:

* `GUI.worldToGrid(pose)`: Converts a pose in world coordinates to grid cell coordinates. Same utility than the old `GUI.rowColumn(pose)` function.
* `GUI.gridToWorld(cell)`: Performs the inverse transformation, from grid cell coordinates to world coordinates. This function was not available previously (not exposed to the user), and it is required in most solutions of the exercise.

Regarding the old `GUI.rowColumn(pose)` function, I propose to deprecate it in favor of the new functions, which have a clearer naming and are less confusing. I still left the `rowColumn()` available to avoid breaking previous solutions.

### DIFF
--- a/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
+++ b/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
@@ -46,13 +46,13 @@ class Map:
 		gridY = (worldX + self.worldWidth/2) * self.gridWidth / self.worldWidth
 		gridX = int(np.clip(gridX, 0, self.gridWidth))
 		gridY = int(np.clip(gridY, 0, self.gridHeight))
-		return [gridX, gridY]
+		return (gridX, gridY)
 
 	def gridToWorld(self, gridX, gridY):
 		""" Transform a map cell to world coordinates """
 		worldX = gridY * self.worldHeight / self.gridHeight - self.worldHeight/2
 		worldY = gridX * self.worldWidth / self.gridWidth - self.worldWidth/2
-		return [worldX, worldY]
+		return (worldX, worldY)
 
 	def RTFormula(self):
 		RTz = self.RTz(pi/2, 50, 70, 0)
@@ -64,14 +64,14 @@ class Map:
 
 	def setGridVal(self, x, y, val):
 		self.grid[y][x] = val
-		
+
 	def getTaxiCoordinates(self):
 		""" Return the taxi pose in map coordinates """
 		pose = self.pose3d()
-		return worldToGrid(pose.x, pose.y)
+		return self.worldToGrid(pose.x, pose.y)
 
 	def rowColumn(self, pose):
-		# Deprecated. Use gridToWorld()
+		# Deprecated. Use worldToGrid()
 		x = pose[0]
 		y = pose[1]
 		#print("x : {} , y : {}".format(x,y))
--- a/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
+++ b/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
@@ -6,11 +6,13 @@ import matplotlib.pyplot as plt
 
 class Map:
 	def __init__(self, pose3d):
+		# The world scenario is placed on 0,0. It has a size of 500x500
+		# The coordinates of the map are from 0 to 400, being 0,0 the top left corner
 		self.pose3d = pose3d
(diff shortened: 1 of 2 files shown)


exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
```diff
@@ -70,24 +66,12 @@ def setGridVal(self, x, y, val):
...
-		x = x * 400
-		y = y + 250
-		y = y / 500
-		y = y * 400
-		#print("x : {} , y : {}".format(x,y))
-		return y, x
+		return worldToGrid(pose.x, pose.y)
 
 	def rowColumn(self, pose):
+		# Deprecated. Use gridToWorld()
```
javizqh: Change to worldToGrid()

exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
```diff
@@ -70,24 +66,12 @@ def setGridVal(self, x, y, val):
...
-		# The coordinates of the map are from 0 to 400, being 0,0 the top left corner
-		x = x + 250
-		x = x / 500
-		x = x * 400
-		y = y + 250
-		y = y / 500
-		y = y * 400
-		#print("x : {} , y : {}".format(x,y))
-		return y, x
+		return worldToGrid(pose.x, pose.y)
```
javizqh: Change to
```python
 return self.worldToGrid(pose.x, pose.y)
```

exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
```diff
@@ -39,25 +41,19 @@ def RTGridWorld(self):
...
-		final_poses = self.RTWorldGrid() * orig_poses
-		gridX = int(final_poses.flat[0])
-		gridY = int(final_poses.flat[1])
-		return (gridX, gridY)
+		""" Transform a world point to map cell coordinates """
+		gridX = (worldY + self.worldHeight/2) * self.gridHeight / self.worldHeight
+		gridY = (worldX + self.worldWidth/2) * self.gridWidth / self.worldWidth
+		gridX = int(np.clip(gridX, 0, self.gridWidth))
+		gridY = int(np.clip(gridY, 0, self.gridHeight))
+		return [gridX, gridY]
```
javizqh: Return tuple not array

exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
```diff
@@ -39,25 +41,19 @@ def RTGridWorld(self):
...
-		worldX = gridX - 250
-		gridY = gridY / 400
-		gridY = gridY * 500
-		worldY = gridY - 250		
-		return (worldX, worldY)
-		
+		""" Transform a map cell to world coordinates """
+		worldX = gridY * self.worldHeight / self.gridHeight - self.worldHeight/2
+		worldY = gridX * self.worldWidth / self.gridWidth - self.worldWidth/2
+		return [worldX, worldY]
```
javizqh: Return tuple not array

javizqh (changes requested):
Please change the lines requested

exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/map.py
ghost: The original `rowColumn()` function returned an array, and this is why the return type is an array too here.
Is there a reason to prefer the tuple format?

javizqh: For the GUI it is necessary that both of this functions return a tuple. You can change the deprecated rowColumn() function in the GUI to return a list instead of the tuple

ghost: Sorry for the miss, I just saw the problem inside the `update_gui()` method when building the message payload.
Fixing and pushing ASAP.

Will still keep the old rowColumn returning a list instead of a tuple just in case.

ghost:
@javizqh thanks for the review! I just sent the updates, let me know if anything else is missing.

javizqh (approved):
Works fine

PR #2721, opened 2024-09-13
Add watchdog dependency

javizqh:
Needed to test the watchdog for the RAM used by bt-studio


javizqh:
Creating new one without conflicts

PR #2798, opened 2024-10-17
Updating amazon warehouse frontend

javizqh:


javizqh:
Missing resize and draw target

javizqh:
Missing changing vehicle model to fit the ackermann one

javizqh:
If possible test in Unibotics before merging. The names of the universes may cause error

dduro2020:
It works perfectly, but I think it would be a good idea to modify the image of the robot on the map for when the robot stands up, maybe by putting a black circle in the centre or some other way that is visually understandable.

javizqh:
I will create a new issue

PR #2800, opened 2024-10-17
Migrating db to RI

### LINKED ISSUE #2831: DB refactoring
- [ ] instead of using SQLite for the Excercise DB we are going to migrate to PostgreSQL.
- [ ] the Universe DB should not be in this repo, but in [RoboticsInfrastructure](https://github.com/JdeRobot/RoboticsInfrastructure).

The Exercise DB should include all the available universes for each RA exercise, and also the default one for each exercise.

### LINKED ISSUE #2833: Remove Ros1 duplicate exercises
Remove Ros1 exercises of already existing ones in Ros2 and also remove old exercises that have a new implementation

### LINKED ISSUE #2834: [Monte Carlo Laser Loc] bugs in poseToMap
The method `poseToMap` from the [Monte Carlo Laser Localization exercise](https://github.com/JdeRobot/RoboticsAcademy/blob/608da5a8375e40d2061f26f1243fb61dcbd88054/exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/GUI.py#L87) returns floating numbers, which are not valid map coordinates. One would expect to receive the indices of the map cell corresponding to the given pose.

Moreover, the methods `poseToMap` and `mapToPose` are not symmetrical: If one creates a cell in map coordinates, converts it to a pose, and then converts it back to map coordinates, the values are different. You can see an example in this image:

Screenshot from 2024-11-04 17-09-59

From the [code related to this functions](https://github.com/JdeRobot/RoboticsAcademy/blob/608da5a8375e40d2061f26f1243fb61dcbd88054/exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/GUI.py#L87-L97), I see that the offset applied when doing the transformation is not the same in both directions:
```Python
    def poseToMap(self, x_prime, y_prime, yaw_prime):
        x = 101.1 * ( 4.2 + y_prime)
        y = 101.1  * ( 5.7 - x_prime)
        yaw = yaw_prime + math.pi/2
        return [x, y, yaw]


    def mapToPose(self, map_x, map_y, map_yaw):
        x = (map_y - 576.27) / -101.1
        y = (map_x - 425.25) /  101.1
        yaw = map_yaw - math.pi/2
        return [x, y, yaw]
```

I don't know where do these numbers come from, so I don't know which one is the correct offset. Could you please fix this?

javizqh:
Migrating universes db to RI


jmplaza:
Related to #2831

javizqh:
Missing Admin page and a way of waiting for the db to be loaded to launch the second docker

javizqh:
Missing autosave when closing docker

PR #2812, opened 2024-10-21
Update 3d reconstruction frontend

javizqh:
Update 3d reconstruction frontend


javizqh:
Just fixed broken css. I think changing the canvas in this exercise  does not make sense because the purpose is just to draw points that are static

dduro2020 (approved):

PR #2820, opened 2024-10-24
Removing Target reset in Global Nav

### LINKED ISSUE #2817: [Global Navigation] Clicked target is deleted after starting exercise
When clicking in the "play" button to run the exercise, the target that was marked in the GUI is removed (cleared). In previous versions the target stayed in the GUI after clicking "play".

It is ok to clear everything after clicking the "reset" button, but removing the target after clicking "play" is a bit painful for the user, because it requires setting a target while the exercise is running. Before, one could set the target and then run the exercise, which was much easier IMHO.

### LINKED ISSUE #2818: [Global Navigation] Unwanted debug trace in console
The console for the "Global Navigation" exercise shows an unwanted debug trace every time the user clicks the GUI to select a target ([link to line in the code](https://github.com/JdeRobot/RoboticsAcademy/blob/38228c43b8eeb284d0550aa1cbcf500a0969b3b2/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/GUI.py#L47)).

This print is not generated by user code and might generate confusion.

javizqh:
Removing Target reset in Global Nav

### DIFF
--- a/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/GUI.py
+++ b/exercises/static/exercises/global_navigation_newmanager/python_template/ros2_humble/GUI.py
@@ -44,7 +44,6 @@ class GUI(MeasuringThreadingGUI):
             self.mapXY = data
             x, y = self.mapXY
             self.worldXY = self.map.gridToWorld(x, y)
-            print(f"World : {self.worldXY}")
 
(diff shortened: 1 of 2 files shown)


dduro2020:
It works perfectly, now the targets are not reset when pressing play and no traces are printed that have not been written by the user.

PR #2884, opened 2024-11-21
Update map.py

saikarna913:
optimising the setSonarValues class


jmplaza:
Hi @saikarna913 , 

have you really tested it? This exercise is deprecated, and needs a complete refactoring. It will change even the simulator, from STDR to Gazebo. This change is currently useless, sorry. And without testing it can't be merged.

PR #2885, opened 2024-11-21
Fix montecarlo laser loc bugs

### LINKED ISSUE #2882: [Monte Carlo Laser Loc] Problem with angle in gazebo <--> map coordinates
Related to #2834.
The conversion functions `GUI.poseToMap` and `GUI.mapToPose` are not transforming the angle properly, it has an offset of 180º.

The converted (x, y) coordinates are fine now, but the angle in map coordinates is wrong (it is not aligned with the X axis).
I think it would be fixed by substracting `pi/2` [here](https://github.com/JdeRobot/RoboticsAcademy/blob/54c7f4fca512f640c6ec0c18839a89ffad2542d5/exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/GUI.py#L90), instead of adding it (and the inverse operation in the other function).

### LINKED ISSUE #2883: [Monte Carlo Laser Loc] Problem with GUI.showPosition
The method `GUI.showPosition` is used to display the "predicted" robot pose. I found 2 issues with this method:

1. The pose is not cleared in the GUI after resetting the exercise. Actually, this problem also happens with the displayed particles (they are not cleared with the reset button).
2. The displayed angle is wrong. The following image shows a simple example where I took the robot's pose from the HAL module and just displayed it with the `GUI.showPosition` method:

show_position_issue

javizqh:
Fix montecarlo laser loc bugs

### DIFF
--- a/exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/GUI.py
+++ b/exercises/static/exercises/montecarlo_laser_loc/python_template/ros2_humble/GUI.py
@@ -87,13 +87,13 @@ class GUI(MeasuringThreadingGUI):
     def poseToMap(self, x_prime, y_prime, yaw_prime):
         x = 101.1 * ( 4.2 + y_prime)
         y = 101.1  * ( 5.7 - x_prime)
-        yaw = yaw_prime + math.pi/2
+        yaw = yaw_prime - math.pi/2
         return [round(x), round(y), yaw]
 
     def mapToPose(self, map_x, map_y, map_yaw):
         x = (map_y - 576.27) / -101.1
         y = (map_x - 424.62) /  101.1
-        yaw = map_yaw - math.pi/2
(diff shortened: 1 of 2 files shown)


javizqh:
@franmore-urjc  Can you test if everything is solved

ghost:
Sorry, but I am afraid I can't. I can't run the local RA version anymore.

I used to be able to run the dev docker compose without having to build the react frontend, but now I cannot do it, and I really don't want to install nvm and yarn in my system for other reasons.

If it helps, the changes look good to me, but I cannot test them.

javizqh:
Okay. For me this solve the issues, so I Will merge It. If the problems still remain, open the issues again

javizqh:
@franmore-urjc In a couple hours there will be a robotics-academy docker image with the tag "test" so that you can confirm if this has been solved

PR #2886, opened 2024-11-21
Adding more info to Monaco

### LINKED ISSUE #2873: [Monaco Editor] The editor doesn't have a way to let the users know of hidden features
The Monaco editor has ctrl + scroll up/down mapped to increase/decrease the size of text. There should be some way to let users know of this feature. 

The old editor had +,- buttons on the editor.

image

javizqh:


javizqh:
Works fine, but I think it will be better to change the zoom icons for just a + and a - to be more readable in smaller screens.

codezerro:
@javizqh icon updated. take a look.

react_frontend/src/index.css
```diff
@@ -1,17 +1,17 @@
...
 code {
-  font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
+  font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
 }
 
 .app {
   margin-bottom: 28px;
-}
+}
```
codezerro: @javizqh i was tring to add some css effect when click on buttons. but removed css because it's not working very well in a small button.

javizqh:
Now it is perfect

javizqh:
Thanks for your amazing work

PR #2891, opened 2024-11-22
[Discontinued] Add Marker visual loc exercise

javizqh:
Add Marker visual loc exercise


javizqh:
Do not close until rebase is done

PR #2956, opened 2025-02-12
Fixing RTF value in harmonic

### LINKED ISSUE #2917: RTF value does not work with Gazebo Harmonic
RTF value does not work with Gazebo Harmonic. It displays 

`RTF: Use 'gz help <command>' to print help for a command.`

dduro2020:


javizqh:
Instead of changing a gui_interface I think it is better to create a new one named measuring_threading_gui_harmonic.py and use that one in the exercises that have been migrated to Harmonic

javizqh (approved):

Issue #2911, opened 2024-12-13
Marker Visual Loc Robot goes to another pose on reset or play


javizqh:
Marker Visual Loc Robot goes to another pose on reset or play.

It goes to the center of the image for a couple seconds, then the position corrects itself

AshishRamesh:
fixed this in #2995

PR #2995, opened 2025-03-12
Fixed Marker Visual Loc Robot goes to another pose on reset / play as mentioned in #2911

AshishRamesh:


javizqh (changes requested):
Values must not be hardcoded

AshishRamesh:
removed hard coded values

javizqh (changes requested):
The coordinantes 0,0 are valid

AshishRamesh:
ive fixed it , now uses the time stamp instead so avoids filtering any valid coordinates.
Side note while looking for possible solutions i noticed that most service robot exercises use hardcoded values (init coords and start coords) to get around this. Should i rectify them in another pr ??

AshishRamesh:
Hi , any changes to be made for this ?

PR #3037, opened 2025-04-08
Visualization error in laser mapping

Adarshgupta1127:


javizqh:
This issue is already being worked on and this fix does affect the user experience instead of only being reset in the browser.

Next time make sure to ask if an issue is being worked on or not.

PR #3038, opened 2025-04-08
Restore exercise human detection

Adarshgupta1127:


javizqh (changes requested):
Sorry, but this is not how the execution works. You may want to take a look at how the other exercises work and the [RAM](https://github.com/JdeRobot/RoboticsApplicationManager).

Also make sure to only open a PR if it works in your setup, because this will never work.

PR #3199, opened 2025-08-19
Car junction exercise

aastha51551:
This is the updated car junction exercise built on Gazebo Harmonic and ROS2 Humble
I have introduced the following improvements and additions

- Added Python Templates
- Modified react components (front camera view with pose coordinates)
- Added a specific hal interface for lidar PointCloudData
- Updated exercise template
- Added the exercise to the database in PROTOTYPE stage


javizqh:
The simulation is not working currently. Take a look at the database in the humble-devel branch in RoboticsInfrastructure.

Right now it is set as a Gazebo Classic exercise, but it does not work with gz sim either. Please take a look at that

aastha51551:
> The simulation is not working currently. Take a look at the database in the humble-devel branch in RoboticsInfrastructure.
> 
> Right now it is set as a Gazebo Classic exercise, but it does not work with gz sim either. Please take a look at that

I am looking into it. I will update you soon.

PR #3332, opened 2025-11-24
Save search bar and filter state using sessionStorage (fixes #3257)

### LINKED ISSUE #3257: Filter state is not saved
The filter state is not saved

(Browser: Google Chrome)

https://github.com/user-attachments/assets/9ceb2403-d65b-4c1b-9335-d7a21c32795c

yujiiroo:
This PR implements sessionStorage-based persistence for the search bar and filter menu (fixes #3257).

### What’s improved
- Filters (Tags / Description / Status) now remain active when:
  - Reloading the exercises page
  - Navigating into an exercise and back
- Search bar text is also restored during the same browser session
- Both filter state and search text reset automatically when the browser/tab is closed

### DIFF
--- a/react_frontend/src/components/buttons/Search.tsx
+++ b/react_frontend/src/components/buttons/Search.tsx
@@ -13,6 +13,7 @@ import {
 import { useHomepage } from "Contexts/HomepageContext";
 import { useState, useEffect } from "react";
 import { useAcademyTheme } from "Contexts/AcademyThemeContext";
+import { Filters } from "src/types/exercises";
 
 // SessionStorage keys
 const FILTER_STORAGE_KEY = "ra_home_filters_v1";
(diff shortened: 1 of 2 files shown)


javizqh:
It saves the filter, but the filtering does not work.

Here is how it should work:

https://github.com/user-attachments/assets/d64e4ed8-ed10-4717-89f0-473069b1803c

Here is your solution:

https://github.com/user-attachments/assets/ebbc25a8-8194-4e08-b817-1db72e3ca676

yujiiroo:
Hey @javizqh ,
 Thanks for pointing out the issue.
I’ve pushed a fix that uses setFilterItemsList to correctly apply the state on load.
Filtering should now match the expected behavior.
Please review when you get a chance.

javizqh (approved):

PR #3402, opened 2026-01-22
Fix webgui desync and use consumer rate of the browser

### LINKED ISSUE #3330: WebGUI frontend desync with Harmonic sim

javizqh:


javizqh:
Does not load properly on entering an exercise

PR #3416, opened 2026-02-01
fix(core): enforce strict cache revalidation to resolve stale UI loading

### LINKED ISSUE #3108: Cookies and local storage are stored in the browser and may cause issues such as failure to launch exercises.
After not using RA for a few months, upon launching, I saw an older UI, which was unresponsive and did not trigger the exercise launch.

Image

A workaround is to clear the browser cache or use incognito mode.

Can we force the browser in some way to clear RA cache on launch, or not store the cache at all (not sure about this one)

karttikjangid:
Fixes #3108

**Description**
This PR resolves the "stale UI" crashes by enforcing a strict `no-cache` policy at the Django view layer.

**The Issue**
Users frequently encounter unresponsive UIs because browsers cache the `index.html` entry point, which references outdated/deleted JS bundles.
* **Investigation:** Static analysis of `react_frontend/views.py` revealed the `academy` view lacked cache-control decorators.
* **Verification:** Production headers currently show no `Cache-Control`, allowing heuristic caching.

**The Fix**
Applied the `@never_cache` decorator to the `academy` view in `react_frontend/views.py`.
* This forces the server to send `Cache-Control: max-age=0, no-cache, no-store, must-revalidate`.
* Ensures the browser always fetches the latest `index.html`, which correctly points to the new hashed JS bundles (verified in build logs as `index.[hash].js`).

**Verification**
* Validated that frontend builds use content hashing.
* Confirmed via static analysis that `urls.py` routes the root path to the modified `academy` view.


javizqh:
Good job, the only remaining problem is that the cookies and local storage are still stored. I will merge it but the issue will remain open.

PR #3421, opened 2026-02-02
Add YouTube frame extractor tool for Computer Vision exercises

### LINKED ISSUE #3058: Basic Computer Vision extension for getting frames from videos at YouTube

Mahaboobunnisa123:
Hi! This PR adds a tool to extract frames from YouTube videos.

I noticed issue #3058 mentioned needing this feature for CV exercises, so I implemented a simple Python script that: 
- Takes a YouTube URL and extracts frames from the video
- Lets you configure how many frames to extract and the interval between them
- Saves frames as JPG images in a folder
- Has a simple command-line interface

The tool uses yt-dlp to get the video stream and OpenCV to extract frames. I added a README with usage examples and installation instructions. This should be helpful for students who want to test their CV algorithms on real videos without downloading huge files.

**Dependencies needed:**
- yt-dlp
- opencv - python

Let me know if you'd like me to change anything or add more features!
Fixes #3058

### DIFF
--- a/scripts/tools/README.md
+++ b/scripts/tools/README.md
@@ -1,0 +1,53 @@
+# YouTube Frame Extractor
+Extract frames from YouTube videos for Computer Vision exercises in RoboticsAcademy.
+
+## Purpose
+This tool allows students to:
+- Extract frames from YouTube videos containing robotics/computer vision content
+- Use real-world video data for testing CV algorithms
+- Avoid downloading entire videos (saves disk space)
+
+## Installation
+Install required dependencies:
+```bash
+pip install yt-dlp opencv-python
+
+## Usage
+Basic Usage
+python youtube_frame_extractor.py "https://www.youtube.com/watch?v=VIDEO_ID"
+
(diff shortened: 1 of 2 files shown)


javizqh:
This does not work. We want to use a Youtube video the same way as we have the local video tool. This is just a python script

Mahaboobunnisa123:
Thank you for the quick feedback @javizqh! 
I understand now - you want this integrated with the existing platform similar to the local video tool, not a standalone script. I'll study how the local video tool is implemented and redesign this to follow the same pattern. Could you point me to the specific files for the local video tool so I can understand the expected structure? I'll update this PR with the proper integration once I understand the architecture.

Mahaboobunnisa123:
Thank you @javizqh for the clarification! I now understand - this needs to be a React component integrated into the frontend, not a backend script. I'll study the local video tool implementation and the WebRTC camera pattern to understand the correct architecture. I'll work on it and come back with the right solution.

PR #3434, opened 2026-02-04
Reduce number of failed connections with Robotics Backend

javizqh:
Needs more testing because on exercise exit after connecting to the Robotics Backend it stills tries to connect 2 more times and fails. This must be 0


javizqh:
The 2 traces of failed to connect after exiting an exercise are traces when closing the websocket

PR #3471, opened 2026-02-13
Fix DeprecationWarning: Replace 'imp' with 'types' in Laser Loc exercise

ayham-engineering:
The imp module was removed in Python 3.12, causing exercises/laser_loc/exercise.py to crash on startup. This PR replaces imp.new_module with types.ModuleType, restoring compatibility with modern Python versions while maintaining functionality. Tested on Python 3.14

### DIFF
--- a/exercises/laser_loc/exercise.py
+++ b/exercises/laser_loc/exercise.py
@@ -12,7 +12,7 @@ from datetime import datetime
 import re
 import json
 import traceback
-import imp
+import types
 import importlib
 import rospy
 from std_srvs.srv import Empty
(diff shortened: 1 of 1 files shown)


javizqh:
We have not updated the Python version since a couple of years. I do not know how you updated It but we only support what is inside the Robotics Backend or docker enviroment.

javizqh:
Also this exercise is not supported

ayham-engineering:
Alright, understood
My bad for testing against a Python 3.12 environment instead of docker

I will focus on the supported exercises using the official docker environment

Thanks for clarification!

PR #3473, opened 2026-02-13
removed redundant react import 

andersendsa:
React is imported twice in this file – once for the hooks and once as a default import
This PR cleans up duplicate React imports in multiple files by combining them into a single import
closes issue :  #3472


javizqh:
This does not fix any duplicate import. You removed the warning that makes obligatory to import react on all react files.

javizqh:
This does not fix any duplicate import. You removed the warning that makes obligatory to import react on all react files.

andersendsa:
thanks @javizqh for the feedback  
i will take it into account for my  future contributions

PR #3481, opened 2026-02-19
Updated the Docs by adding run methord for RADI

Sarvesh-Mishra1981:
The method to run the custom RADI was not mentioned clearly. I have added detailed instructions describing how to execute it using the run_academy.sh script.
This will make it easier for developers who want to run a custom RADI configuration, as they can now directly use the run_academy.sh script without confusion.

Screenshot 2026-02-20 at 1 48 18 AM


javizqh:
run_academy is not meant to be used with custom RADIs, as those are part of the developer experience and should be used in the same way as stated in the developer guide

PR #3483, opened 2026-02-20
New ram as python package

javizqh:


JitinSaxenaa:
Hi @javizqh 

After this PR, Docker Compose still tries to run
/RoboticsApplicationManager/manager/manager/manager.py, which no longer exists since RAM is now a Python package.

Getting:
python3: can't open file ... manager.py

Should the entrypoint or the docs instructions be updated?

javizqh:
Please update all of the others files.

javizqh:
Also docker compose does not run anything

PR #3484, opened 2026-02-20
Fixed build issue on macOs

Dev-sh21:
hi @jmplaza @javizqh, sir

This PR fixes a Docker build issue on macOS where the colcon build step was failing with a “write jobserver: Bad file descriptor” error while building the as2_core package. The problem was due to the default parallel build setup not working properly in the macOS Docker environment. To fix it, I forced single-threaded compilation by adding export MAKEFLAGS="-j1" before each colcon build command and updated the commands to use --executor sequential and --parallel-workers 1. These changes were made in scripts/RADI/Dockerfile.dependencies_humble, and now the Docker image builds successfully on macOS without affecting any other functionality.

### DIFF
--- a/scripts/RADI/Dockerfile.dependencies_humble
+++ b/scripts/RADI/Dockerfile.dependencies_humble
@@ -296,7 +296,8 @@ RUN apt-get update && \
       && rm -rf /var/lib/apt/lists/*
 
 RUN . /opt/ros/$ROS_DISTRO/setup.sh && \
-    colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release
+    export MAKEFLAGS="-j1" && \
+    colcon build --parallel-workers 1 --executor sequential --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release
 
 # Install CycloneDDS RMW for ROS 2 Humble to fix cycle time issues in humble-moveit (temporary fix)    
 RUN echo 'export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp' >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


javizqh (changes requested):
Please make it only for Mac, not for every architecture. This slows the compiling for all other architectures

Dev-sh21:
thankyou @javizqh sir, for your review

i definitely dont want to slow down the builds for everyone else,
since the build runs inside a Linux container, docker doesnt natively expose the host OS without passing a custom build argument.As per my research and knowledge, i have a couple of options to restrict this to macOS

1) use TARGETARCH as a proxy for Apple Silicon Macs, assuming the issue predominantly affects m-series macs, we can target arm64-
```
ARG TARGETARCH
RUN if [ "$TARGETARCH" = "arm64" ]; then \
      export MAKEFLAGS="-j1" && colcon build --executor sequential --parallel-workers 1; \
    else \
      colcon build; \
    fi

```

2) use a custom build argument- we can introduce an ARG MAC_BUILD=false in the Dockerfile, and mac users would need to run docker build --build-arg MAC_BUILD=true.

3) check for docker desktop's kernel (linuxkit) -> docker desktop on mac typically uses the linuxkit kernel, so we could check uname -r for "linuxkit".


which approach do you think fits best with the repository's practices? 
I'll update the pr right away once you confirm.

javizqh:
Use the one you like the most. I cannot test it on Mac and it is part of your contribution to find the best approach.

Dev-sh21:
@javizqh sir,

I've implemented Option 1 using TARGETARCH, now the sequential build workaround (-j1) only runs when building on arm64 (Apple Silicon). Standard amd64 nodes (like your CI runners) will automatically bypass this and continue using the normal, fully-parallelized colcon build

This prevents slowing down the builds for everyone else while fixing the macOS jobserver issue. let me know if this looks good to merge

Dev-sh21:
@javizqh sir, do i have to make any more changes?

Dev-sh21:
hi @javizqh sir 

is there any change required in this solution?

PR #3509, opened 2026-03-01
Fix - Prevent Follow Line map rendering before connection

JitinSaxenaa:
## Bug Description :

In the Follow Line exercise, the circuit map was rendered 
immediately when the page loaded, even before the exercise was connected.

## Fix :

The circuit overlay is now conditionally rendered only when the 
connection is done. Since the image is received only after 
connection and RUNNING state, this ensures the map is displayed 
at the correct time.

## Files Modified :

- exercises/follow_line/frontend/WebGUI.tsx

### DIFF
--- a/exercises/follow_line/frontend/WebGUI.tsx
+++ b/exercises/follow_line/frontend/WebGUI.tsx
@@ -20,7 +20,7 @@ const WebGUI = () => {
   const [image, setImage] = useState<string | undefined>(undefined);
   const [lapTime, setLapTime] = useState<string | undefined>(undefined);
   const [carPose, setCarPose] = useState<number[] | undefined>(undefined);
-  const [circuitImg, setCircuitImg] = useState(defaultCircuit);
+  const [circuitImg, setCircuitImg] = useState<string | undefined>(undefined);
   const [manager, setManager] = useState(exerciseContext.manager);
 
   useEffect(() => {
@@ -100,7 +100,7 @@ const WebGUI = () => {
           {lapTime} s
         </label>
       )}
-      {image && (
+      {circuitImg && (
         <div className="overlay" id="circuit-aerial">
           <img src={circuitImg} alt="" id="circuit-img" />
           {carPose && (
--- a/exercises/follow_line/frontend/WebGUI.tsx
+++ b/exercises/follow_line/frontend/WebGUI.tsx
@@ -100,15 +100,17 @@ const WebGUI = () => {
           {lapTime} s
         </label>
       )}
-      <div className="overlay" id="circuit-aerial">
-        <img src={circuitImg} alt="" id="circuit-img" />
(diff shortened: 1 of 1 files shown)


javizqh:
This would only show the map if the user sends an image. The circuit must be independant to any user actions

javizqh (changes requested):
Please fix what is said in the comment

JitinSaxenaa:
Hi @javizqh 
Are the changes fine now ?

javizqh (changes requested):
It is better to define the circuitImg as undefined in line 23 and just change circuitReady on line 106 to  circuitImg. It is preferable to have less react states

exercises/follow_line/frontend/WebGUI.tsx
```diff
@@ -80,13 +74,14 @@ const WebGUI = () => {
...
     if (state === states.RUNNING) {
       manager.send("gui", "startLap");
+      updateCircuit(manager.getUniverse());
     } else if (state === states.PAUSED) {
       manager.send("gui", "pause");
     } else if (state === states.TOOLS_READY) {
       setCarPose(undefined);
       setLapTime(undefined);
       setImage(undefined);
-      updateCircuit(manager.getUniverse());
```
javizqh, line 89: Please revert this change. This selects the circuit when the world is ready and does not need to wait until is launched

exercises/follow_line/frontend/WebGUI.tsx
```diff
@@ -80,13 +74,14 @@ const WebGUI = () => {
 
     if (state === states.RUNNING) {
       manager.send("gui", "startLap");
+      updateCircuit(manager.getUniverse());
```
javizqh: This should be done in the tools_ready state

javizqh (changes requested):

exercises/follow_line/frontend/WebGUI.tsx
JitinSaxenaa: Should the map be displayed just after connecting the exercise or after clicking the play button ?

javizqh: It must be displayed when the circuit is selected (when the sim is show)

exercises/follow_line/frontend/WebGUI.tsx
javizqh: This needs to be removed. It does not make any sense as it only triggers an unneeded rerender

javizqh (changes requested):

javizqh:
Please read what I say in the comments and try executing your changes. I have already told you multiple times what to remove and you keep changing what makes the circuit appear on screen. If not I will close this PR and modify It myself

javizqh (approved):

PR #3549, opened 2026-03-10
Fix StyledHeaderButton props using styled-components transient props

deeepakgm-hw:
## Summary

This PR updates the `StyledHeaderButton` component to use **styled-components transient props** to prevent custom styling props from being passed to the DOM.

## Changes

* Converted `bgColor`, `hoverColor`, and `roundness` props to transient props:

  * `$bgColor`
  * `$hoverColor`
  * `$roundness`
* Updated `PlayPause.tsx` to use the new transient props.

## Reason

Previously, props like `bgColor` and `hoverColor` were forwarded to the underlying HTML `<button>` element, which could cause React warnings such as:

> React does not recognize the 'hoverColor' prop on a DOM element.

Using styled-components transient props ensures these styling props are handled internally and not passed to the DOM.

## Impact

* Removes potential React DOM warnings.
* Improves code quality and compatibility with styled-components best practices.

## Files Modified

* `react_frontend/src/styles/headers/HeaderMenu.styles.ts`
* `react_frontend/src/components/buttons/PlayPause.tsx`

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -35,3 +35,21 @@ sudo systemctl disable postgresql
 ```
 
 After this follow the steps provided in [official installation instructions](https://jderobot.github.io/RoboticsAcademy/user_guide/#installation).
+
+## Docker containers exit with status 255
+
+Sometimes the containers `developer-container` and `universe_db` may stop with status 255.
+
+To restart the environment:
+
+```bash
+cd compose_cfg
+docker compose -f dev_humble_cpu.yaml up
+```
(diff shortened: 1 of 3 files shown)


javizqh:
This PR has duplicated code, breaks the play function and also has code not related to it

deeepakgm-hw:
Thanks for reviewing the PR and for the feedback.
I understand the issue with duplicated/unrelated code.

I will look for an existing issue or open one before making further changes so the contribution aligns better with the project needs.

PR #3550, opened 2026-03-10
Replace unsafe eval() with json.loads() for parsing exercise tags

deeepakgm-hw:
This PR replaces unsafe usage of eval() with json.loads() when parsing exercise tags.

Using eval() on database values can execute arbitrary code and is considered unsafe.
json.loads() safely parses the stored JSON string and preserves the same behavior.

Changes:
- Replaced eval(project.tags) with json.loads(project.tags)
- Added safe fallback if parsing fails

This improves backend security and follows Python best practices.

### DIFF
--- a/academy/views.py
+++ b/academy/views.py
@@ -74,6 +74,439 @@ def enter_exercise(fal, request):
     for tool in project.tools.all():
         tools.append(tool.name)
 
+    try:
+        tags = json.loads(project.tags)
+    except Exception:
+        tags = []
+
+    info = {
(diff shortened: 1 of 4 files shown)


javizqh:
This PR has duplicated code, does not fix any issue and also has unrelated code

deeepakgm-hw:
Thanks for reviewing the PR and for the feedback.
I understand the issue with duplicated/unrelated code.

I will look for an existing issue or open one before making further changes so the contribution aligns better with the project needs.

PR #3557, opened 2026-03-11
Fix lint error: remove unused id prop from WebGUIContainer

deeepakgm-hw:
This PR removes the unused `id` prop from WebGUIContainer to fix the ESLint warning (`id` is defined but never used).


javizqh:
There is already a PR open solving this issue

PR #3627, opened 2026-03-20
Optimize RoboticsBackend container by removing optional heavy dependencies

kaushikteja26:
This PR proposes an optimization of the RoboticsBackend container by removing optional components that significantly increase image size and introduce dependency issues.

### Key Changes

- Removed Aerostack2 from the default build:
  - Introduces a large dependency tree (ROS, simulation, control stacks)
  - Not required for most RoboticsAcademy exercises
  - Caused build failures (e.g., missing ignition-gazebo dependencies)

- Removed gz_ros2_control:
  - Depends on additional ROS control packages (e.g., controller_manager)
  - Not required for standard usage
  - Caused build instability

### Results

- Image size reduced from ~29.8 GB → ~16.8 GB
- Reduction of ~13 GB
- Improved build stability

### Rationale

The current container includes several advanced components by default, which increase complexity and setup cost for new users.

This change moves towards a more minimal and stable base image, where advanced components can be modularized or included optionally in future improvements.

### DIFF
--- a/scripts/RADI/Dockerfile.dependencies_humble
+++ b/scripts/RADI/Dockerfile.dependencies_humble
@@ -59,31 +59,19 @@ RUN curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o 
 
 # Install Gazebo 11
 RUN add-apt-repository ppa:openrobotics/gazebo11-gz-cli
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get install -y --no-install-recommends \
   gazebo11 \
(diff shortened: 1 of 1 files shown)


javizqh:
You have removed Gazebo Harmonic (the simulator we use for 90% of the exercises), aerostack (we use it in all of the drones exercises) and gz-control (we use it in the 2 industrial exercises).

With your propose changes we have a total of 0 exercises available in Robotics Academy .

You have not tested any of the changes because if you did that you would not say that nonesense about having more stability and that you only removed optional components

PR #3631, opened 2026-03-20
Clarify C++ execution model

JitinSaxenaa:
### Description

Clarifies the execution model for C++ exercises.

Adds a note explaining that users should not define their own `main()` and must implement `void exercise()`, as the framework already provides `main.cpp`.

### Why

This is a common source of confusion and often leads to errors like multiple `main()` definitions or missing `exercise()`.

### Changes

- Added note under HAL-based C++ section
- Included minimal example


javizqh:
Already fixed in https://github.com/JdeRobot/RoboticsAcademy/pull/3632

Issue #2239, opened 2023-09-13
Optimization of miniRADI docker image size


jmplaza:
Currently it is very large, 4.4GB compressed. It includes all the dependencies of robotics applications, but maybe its size can be reduced somehow.

pariaspe:
Some related issues with recent related work:
- https://github.com/JdeRobot/RoboticsInfrastructure/issues/276 
- https://github.com/JdeRobot/RoboticsInfrastructure/issues/315
- https://github.com/JdeRobot/RoboticsInfrastructure/issues/277
- https://github.com/JdeRobot/RoboticsInfrastructure/issues/280
- https://github.com/JdeRobot/RoboticsAcademy/issues/2227

Surajpatra700:
Sir @jmplaza i tried to build docker image of "Dockerfile.dependencies_humble" file using command "sudo docker build -t mini_radi ./Dockerfile.dependencies_humble ." it give me this error:
![Docker_Error](https://github.com/JdeRobot/RoboticsAcademy/assets/107884058/b17daeb4-2340-4b2d-8826-4ba41979d9d0)
I searched it over Internet but unable to get a solution that worked for me.

pariaspe:
Hey @Surajpatra700,
We have a script in that folder to build RA docker images, try it instead of building manually via cli. Anyway, you are using wrongly the command, I would recommend you to check the docker official documentation.
Btw, you shouldn't use `sudo` before docker commands. Take a look to that too.

jmplaza:
https://hub.docker.com/r/jderobot/robotics-academy/tags

Sarvesh-Mishra1981:
hello @jmplaza, Can i contribute to this.

jmplaza:
Hi @Sarvesh-Mishra1981 , sure. No permission is required, just work on it and share here your findings. Currently profiling the size of all the robotics dependencies already included in the RADI would be useful. Anyway, for newcomers we strongly recommend the betatesting open issues of RoboticsAcademy.

Dev-sh21:
hi @jmplaza sir,

i noticed RADI installs both Gazebo 11 and Harmonic. Since the migration is nearly done, do we plan to keep both in one monolithic image or split them into separate variants to reduce space?

jmplaza:
> hi [@jmplaza](https://github.com/jmplaza) sir,
> 
[...]

PR #3650, opened 2026-03-24
feat: remove node_modules after React build to reduce image size

xu826Jamin:
## Summary
Removes node_modules from the final Docker image after React frontend 
compilation, eliminating 900 MB of unnecessary build-time dependencies.

## Motivation
The React build step installs 2.74 GB of node_modules but only the 
compiled build/ output is needed at runtime. Django serves only the 
static files - node_modules are never used in production.

Related to: #2239

## Changes
- Chain `rm -rf node_modules` after `yarn run build` in the same RUN layer
- Fix conditional sourcing of `local_setup.bash` (pre-existing issue)
- Add `gz_link_attacher` to `colcon --packages-ignore` (pre-existing build fix)

## Size Impact
Total image Before and After -> 47.2 GB -> 31.1 GB -> 34% (16.1 GB) 
React layer Before and After -> 2.74 GB ->1.84 GB ->33% (900 MB)

## Testing
- [x] Image builds successfully
- [x] docker compose up runs correctly
- [x] React frontend loads in browser
- [x] No functionality regression

### DIFF
--- a/scripts/RADI/Dockerfile.humble
+++ b/scripts/RADI/Dockerfile.humble
@@ -37,15 +37,18 @@ RUN git clone --depth 1 https://github.com/${ACADEMY_OWNER}/RoboticsAcademy.git 
 # Compiling and sourcing the workspace
 WORKDIR /home/ws
 RUN /bin/bash -c "source /home/drones_ws/install/setup.bash"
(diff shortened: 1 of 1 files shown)


javizqh:
This is not an issue and it is irrelevant. Please open PR to fix real problems or current issues

Issue #3111, opened 2025-06-12
Coding inline documentation


pariaspe:

monish0705:
Hi!  I’d like to work on this issue.
Could you please guide me on which files/functions to start documenting?

javizqh:
This documentation is only for the backend and frontend files, not exercise documentation

karttikjangid:
I’m investigating whether this issue still occurs with the current frontend.
I’ll test cache behavior (service workers, cache headers, local storage) and report findings before proposing a fix.

javizqh:
> I’m investigating whether this issue still occurs with the current frontend. I’ll test cache behavior (service workers, cache headers, local storage) and report findings before proposing a fix.

Wrong issue @karttikjangid .

manashree-kalhapure:
Hi! Is the issue still available ?

javizqh:
Yes

GUYUEFANGYUAN8:
Hi @jmplaza and @javizqh , I would like to contribute to #3111.

I have manually traced the backend logic in the humble-devel branch and prepared a focused, documentation-only PR for the following files:

academy/file_access.py

academy/error_handler.py

The documentation aims to clarify the internal "contracts" of the File Abstraction Layer (FAL), specifically:

Security Bounds: Why path validation (e.g., ".." checks) is centralized in the abstract layer.

Container Compatibility: The rationale behind permissive file mode normalization (0o777) after writes to prevent UID conflicts in Docker.

Return Type Oddities: Clarifying that exists() returns int (size/status) rather than bool to support Academy's internal metadata needs.

This PR is strictly limited to Docstrings and Type Hints. No functional behavior is changed. I’ve verified the formatting with pydocstyle to ensure compliance with the project's standards. Hope u can give me some useful suggestions.

javizqh:
This is not a PR, this is an issue

PR #3654, opened 2026-03-25
Docs/backend 3111 

GUYUEFANGYUAN8:
Hi, this PR provides comprehensive inline documentation (Google Style) for core backend modules as part of #3111.

Beyond just adding docstrings, I have manually traced the logic flow to ensure the documentation accurately reflects the system's internal contracts.

Key Technical Contributions:

Type Hint Correction: Fixed a mismatch in the exists() method. The return type was hinted as bool, but the implementation correctly returns int (-1 for missing, 0 for directory, or file size). This ensures better compatibility with static analysis tools like mypy.

Architectural Rationale: Documented the necessity of 0o777 permissions in file_access.py to handle cross-container UID accessibility, which is critical for the Academy's deployment stack.

Decorator Logic: Clarified the error_wrapper logic in error_handler.py, specifically regarding request-level isolation (FAL instance cloning) and parameter validation for tuples.

Verification:

Validated against flake8 and pydocstyle.
屏幕截图 2026-03-25 094830
屏幕截图 2026-03-25 094230

Verified that the type hint changes remain consistent across the inheritance chain (FAL -> FAL_RA).

No functional logic was changed.

### DIFF
--- a/academy/error_handler.py
+++ b/academy/error_handler.py
@@ -65,7 +65,9 @@ def error_wrapper(type: str, param: list[str | tuple] = []):
 
 
 def check_parameters(request, param: list[str | tuple]):
-    """Validate required request parameters."""
+    """
+    Validate that required fields are present and meet length constraints.
+    """
 
     for p in param:
         min_len = 0
--- a/academy/file_access.py
+++ b/academy/file_access.py
@@ -53,7 +53,7 @@ class FAL(ABC):
         pass
 
     @abstractmethod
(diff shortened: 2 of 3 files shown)


javizqh:
There are already inline comments for all of the functions changed. Please try to focus on the parts that do not have them, not changing what we have.

Before submitting the PR check if there have been similar PRs merged, because there was one adding those comments in the last few days

PR #3669, opened 2026-03-26
fix(hal_interfaces): fix BumperNode subscription overwrite and LaserData comment swap

### LINKED ISSUE #3668: [Bug] BumperNode loses right and center subscriptions due to self.sub being overwritten in loop
## Description

`BumperNode.__init__` in `common/hal_interfaces/hal_interfaces/general/bumper.py` subscribes to three bumper topics inside a loop but stores each subscription in the same `self.sub` attribute:
```python
for i in range(len(self.topics)):
    self.sub = self.create_subscription(
        ContactsState, topics[i], self.callbacks_[i], 10
    )
```

Each iteration overwrites `self.sub` with the new subscription. When the loop finishes, only the last subscription (left bumper) is referenced. The right and center bumper subscriptions have no references and can be silently garbage collected by Python, causing those callbacks to stop firing.

## Fix

Store all subscriptions in a list:
```python
self.subs_ = [
    self.create_subscription(
        ContactsState, topics[i], self.callbacks_[i], 10
    )
    for i in range(len(self.topics))
]
```

## Additional fix

`LaserData.__init__` in `laser.py` has swapped comments on `minRange` and `maxRange` — minor but misleading for students.

## Environment
- RoboticsAcademy `humble-devel` branch

Kunal-Somani:
## Summary

Fixes #3668

## Bug 1 — BumperNode loses right and center subscriptions

`BumperNode.__init__` in `bumper.py` subscribed to three bumper topics in a loop but stored each subscription in the same `self.sub` attribute, overwriting it every iteration.

When the loop finished, only the last subscription (left bumper) was referenced. Right and center bumper subscriptions had no
references and could be silently garbage collected by Python, causing those callbacks to stop firing with no error or warning.

**Fix:** Store all subscriptions in `self.subs_` list so all three references stay alive for the node lifetime.

**# Before (buggy)**
```python
for i in range(len(self.topics)):
    self.sub = self.create_subscription(
        ContactsState, topics[i], self.callbacks_[i], 10
    )
```

**# After (fixed)**
```python
self.subs_ = [
    self.create_subscription(
        ContactsState, topics[i], self.callbacks_[i], 10
    )
    for i in range(len(self.topics))
]
```

## Bug 2 — LaserData minRange/maxRange comments swapped

`LaserData.__init__` in `laser.py` had swapped inline comments:
- `self.minRange` was labelled `# Max Range posible (meters)`
- `self.maxRange` was labelled `# Min Range posible (meters)`

Also fixed typo: `posible` → `possible`

### DIFF
--- a/common/hal_interfaces/hal_interfaces/general/bumper.py
+++ b/common/hal_interfaces/hal_interfaces/general/bumper.py
@@ -58,12 +58,12 @@ class BumperNode(Node):
             self.left_callback,
         ]
 
-        # Subscribe to all the callbacks
-        for i in range(len(self.topics)):
-
-            self.sub = self.create_subscription(
-                ContactsState, topics[i], self.callbacks_[i], 10
-            )
+        # Subscribe to all the callbacks and store in a list
+        # to prevent subscriptions from being garbage collected
+        self.subs_ = [
+            self.create_subscription(ContactsState, topics[i], self.callbacks_[i], 10)
+            for i in range(len(self.topics))
+        ]
 
         # Right, center, left
         self.contact_states_ = [ContactsState() for _ in range(3)]
--- a/common/hal_interfaces/hal_interfaces/general/bumper.py
+++ b/common/hal_interfaces/hal_interfaces/general/bumper.py
@@ -58,12 +58,9 @@ class BumperNode(Node):
             self.left_callback,
         ]
 
-        # Subscribe to all the callbacks and store in a list
-        # to prevent subscriptions from being garbage collected
-        self.subs_ = [
+        # Subscribe to all the callbacks
+        for i in range(len(self.topics)):
             self.create_subscription(ContactsState, topics[i], self.callbacks_[i], 10)
-            for i in range(len(self.topics))
-        ]
 
         # Right, center, left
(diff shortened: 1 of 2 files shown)


common/hal_interfaces/hal_interfaces/general/bumper.py
```diff
@@ -58,12 +58,12 @@ def __init__(self, topics):
             self.left_callback,
         ]
 
-        # Subscribe to all the callbacks
-        for i in range(len(self.topics)):
-
-            self.sub = self.create_subscription(
-                ContactsState, topics[i], self.callbacks_[i], 10
-            )
+        # Subscribe to all the callbacks and store in a list
```
javizqh: Leave it as it was, just remove the `self.sub = ` part, as that variable is not used anywhere

javizqh (changes requested):

common/hal_interfaces/hal_interfaces/general/bumper.py
Kunal-Somani: I've removed the `self.sub =` assignment and kept the original loop structure. Thank you for the clarification. If any more changes are required, just tell me and i will do it right away.

javizqh (approved):

PR #3693, opened 2026-03-28
Add My-name to Contributors list

AYMENsl25:


jmplaza:
Sorry @AYMENsl25 , but more significant and long term contributions on RoboticsAcademy are required before appearing there.

javizqh:
@AYMENsl25 you must contribute before adding your name there.

Also we do not keep a list of all contributors in the Readme because GitHub does keep track of all of them automatically

PR #3725, opened 2026-04-02
Improve power tower inspection docs with structured detection representation

keerthivnair:
This PR improves the power tower inspection exercise documentation by introducing an optional structured representation for detected defects.

The proposed format (label, bounding box, confidence) helps:
- separate detection logic from visualization
- improve debugging and logging
- prepare the pipeline for future integration with deep learning models

This change is purely additive and does not affect the core objective of the exercise.


javizqh:
Hi @keerthivnair. Your changes do not align with the content of the exercise documentation, as It seems to only be related to your solution. Users may implement that part as they want (It is just a simple array)

PR #3747, opened 2026-04-10
Optimize QuerySet count in views.py by using exists() instead of len(all())

andersendsa:


javizqh:
This is wrong because of:

1. Does not follow the contributor guidelines
2. You changed 16 files, but the your "change" should only change 1
3. There is no issue related

andersendsa:
sorry i commited the wrong files i will make sure commit the correct file and create a issue first and then pull the pr

PR #3753, opened 2026-04-10
Optimize Exercise tool fetching with prefetch_related 

### LINKED ISSUE #3752: Optimize Exercise tools retrieval with prefetch_related
The current logic fetches an Exercise and then immediately accesses its related tools without preloading them.

This introduces an unnecessary extra query today and creates an N+1 query risk if the same pattern is later used to load multiple exercises. In that case, each exercise would trigger its own query for tools, causing avoidable database overhead.

The related tools should be preloaded as part of the initial query so everything is fetched in a single efficient database round-trip.
CC: @javizqh

andersendsa:
closes: #3752

### DIFF
--- a/academy/views.py
+++ b/academy/views.py
@@ -79,7 +79,7 @@ def enter_exercise(fal, request):
     Retrieve basic information about an exercise. Only called when entering.
     """
     project_id = request.GET.get("project_id")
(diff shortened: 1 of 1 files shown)


andersendsa:
Hi @javizqh  could u pls let me know if there are any changes required in the pr or it is it good to merge

PR #3769, opened 2026-04-14
Fix N+1 query overhead

### LINKED ISSUE #3768: N+1 query overhead in get_universes_list
### Description:
The original code implemented this by iterating over all universes using Python's built-in sorted() function. Inside the key lambda for sorting, it called .get() on the ExerciseUniverses through-model for each item to check if it was marked as default (is_default).
This logic resulted in a classic N+1 query problem. For every list of size N, the system would execute 1 initial query to get all universes, plus N individual database queries to check the is_default flag. f the database grows in the future, this  will generate an unnecessary, unscalable load directly proportional to the number of universes.

andersendsa:
closes: #3768

### DIFF
--- a/academy/views.py
+++ b/academy/views.py
@@ -405,18 +405,13 @@ def get_universes_list(fal, request):
     project_id = request.GET.get("project")
     project = Exercise.objects.get(exercise_id=project_id)
 
-    universes_list = []
-
-    proj_univs = project.universes.all()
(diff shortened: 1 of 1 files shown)


javizqh:
Breaks https://github.com/JdeRobot/RoboticsAcademy?tab=contributing-ov-file#opening-a-pull-request

PR #3807, opened 2026-05-06
C++ support loc vacuum cleaner

aquintan4:


javizqh (changes requested):
Please remove the map publisher

PR #3809, opened 2026-05-06
cpp support for amazon warehouse

aquintan4:


javizqh:
Do not remove the comments and also write them in english

javizqh (changes requested):
Add the comments from the last commit in English

javizqh (approved):

PR #3830, opened 2026-05-19
power tower inspection update

aquintan4:
Add model fuction for DL applications


exercises/power_tower_inspection/cpp_lib/include/Model.hpp
javizqh, line 1: This file is not needed

javizqh (changes requested):
Please also add the Python file

javizqh (approved):

PR #3832, opened 2026-05-22
Add machine vision exercise in harmonic

Jepear19:


database/exercises/db.sql
```diff
@@ -214,6 +214,8 @@ COPY public.exercises_universes (id, exercise_id, universe_id, is_default) FROM
 67	14	63	False
 68	14	64	False
 69	14	65	False
+70	27	40	False
```
javizqh: Which exercise is this? We have no exercise with id 27

javizqh (changes requested):

Issue #3806, opened 2026-05-05
C++ Support Beta Testing for Robotics Academy Exercises


aquintan4:
Hi,

There is now support for solving several Robotics Academy exercises in C++, and we need people to actually **build solutions while testing the API**.

The goal is not to review the quality of the solutions themselves, but to gather feedback about the current C++ support, the simple API, the tooling, and the documentation.

**Important:** I’m **not interested in your solution code** and you shouldn’t share it here. What matters is the **feedback** you get while trying to use the C++ support.

## Exercises to test

Feedback is welcome for any of the following exercises:

- [Obstacle Avoidance](https://jderobot.github.io/RoboticsAcademy/exercises/AutonomousCars/obstacle_avoidance)
- [Global Navigation](https://jderobot.github.io/RoboticsAcademy/exercises/AutonomousCars/global_navigation/)
- [Laser Mapping](https://jderobot.github.io/RoboticsAcademy/exercises/MobileRobots/laser_mapping)
- [Marker-based Visual Localization](https://jderobot.github.io/RoboticsAcademy/exercises/ComputerVision/marker_visual_loc)
- [MonteCarlo Laser Localization](https://jderobot.github.io/RoboticsAcademy/exercises/MobileRobots/montecarlo_laser_loc/)
- [Follow Line](https://jderobot.github.io/RoboticsAcademy/exercises/AutonomousCars/follow_line/)

You can test one exercise, several of them, or all of them. Any useful feedback is welcome.

## ⚠️ Generating a new RADI until the new version is released ⚠️

```bash
cd RoboticsAcademy/scripts/RADI
./build.sh  # generates a new RADI image named `test`
```

Then, update your `compose_cfg` file, for example `dev_humble_cpu` if using `./scripts/develop_academy.sh` on Ubuntu, to use the new image:

```yaml
robotics-academy:
  image: jderobot/robotics-academy:test
```

This ensures the Robotics Academy backend runs with the newly built data.

## What to do

- Try to implement a solution to any of the exercises listed above in C++.
- Use the **simple API**:
  - `HAL`
  - `WebGUI`
  - `Frequency`
- Check whether the API behaves as expected while solving the exercise.
- Optional: share a video showing your solution working.

Again, the implementation itself is not the focus. The important part is whether the C++ API and tooling allow you to solve the exercise comfortably and correctly.

## What to report

We are interested in issues related to **our API, tooling and documentation**, not bugs in your own solution code.

In particular, please report:

[...]

PR #3871, opened 2026-06-13
Fix mandatory HAL dependency for ROS2 direct usage in cpp templates

POORVANGIKA:
Fixes the issue reported in #3806 where building a pure ROS 2 solution 
(without the Simple API) fails to compile because `#include "HAL.hpp"` 
was unconditionally included at the top of the generated `main.cpp`.

HAL.hpp is now wrapped in a `#ifndef USER_NODE` guard, so it is only 
included when actually needed.

Affected exercises:
- follow_line
- obstacle_avoidance
- global_navigation
- laser_mapping
- marker_visual_loc
- montecarlo_laser_loc

### DIFF
--- a/exercises/follow_line/cpp_template/main.cpp
+++ b/exercises/follow_line/cpp_template/main.cpp
@@ -1,12 +1,13 @@
-#include "HAL.hpp"
 #include "WebGUI.hpp"
+#ifndef USER_NODE
+#include "HAL.hpp"
+#endif
 #include "academy.cpp"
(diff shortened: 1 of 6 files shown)


javizqh:
Missing video of the before and after of the issue that is meant to be fixed

javizqh (changes requested):
Missing video evidence

aquintan4:
Please make sure you have the latest RADI version. This issue was fixed in #3812

PR #3872, opened 2026-06-13
Add drone_cat_mouse exercise (ROS2)

anishk85:
New two-drone chase exercise for Gazebo Sim. The cat (the part the student
writes) chases the mouse, and the mouse (processB, pre-programmed) flies a
square so there's always something to chase.

Wrote a custom HAL for the exercise instead of touching the shared
DroneWrapper - it tracks the other drone's pose. Added the python_template (cat + mouse) and a
frontend, and seeded the exercise/world/universe rows in db.sql so it shows
up and launches from the UI.

How it runs: the python_template files (the student's academy.py plus the
shared commons - HAL, hal_interfaces, gui_interfaces, etc.) get bundled into
a single app.zip and sent over to RAM. RAM drops it in /workspace/code and
unzips it there, so processB/ lands at /workspace/code/processB/ on its own.
It then starts academy.py as agentA (the cat) and, if it finds
processB/academy.py, starts that as agentB (the mouse) - both as managed
processes RAM can play/pause/stop together.


javizqh:
Please resolve conflicts

PR #3900, opened 2026-07-05
[WIP] Multi-stage build for Dockerfile.dependencies_humble

karttikjangid:
This is a full multi-stage rewrite of Dockerfile.dependencies_humble, not
a small tweak — the diff goes from 373 to 635 lines. Opening as Draft
since to showcase the changes i will make in it in upcoming days .
## What changed
Split into two stages:
- `deps-builder`: unchanged from the original file — same commands, same
  order, compiles everything (ROS workspaces, gz_ros2_control, aerostack2).
- `deps-final`: new. Reinstalls only the packages actually needed, copies
  the compiled workspace output wholesale (no subsetting), keeps the full
  ROS build toolchain.

## Why the toolchain stays
Dockerfile.humble compiles CustomRobots/jderobot_drones/Industrial/
common_interfaces_cpp directly on top of this image via its own colcon
build call. A pure-runtime deps-final would break that build. Confirmed
via Slack
## What got trimmed
- tmux/tmuxinator — confirmed zero invocations via a full boot/launch
  call-graph trace (dead code, not just unused-by-assumption).
- postgresql-18 -> libpq5 — confirmed the DB server binary is never
  started in this image, only the client library psycopg2 links against.

## Verified
- Full ldd sweep, zero unresolved libraries, across both workspaces.
- Django/DB/RAM-websocket/noVNC endpoint tests via docker compose, all
  passing.
- Image size: 11.9GB -> 10.7GB, measured same-day against a freshly
  rebuilt copy of the original file (not an old/stale number).

## Out of scope for this PR
- Dockerfile.humble's own multi-stage split — planned as a follow-up
  once this is merged or another approach would be to change both of them test them completely giving some more time and then merge both pr's.
- An unrelated issue found while testing exercise loading through
  docker compose (two database init SQL files had been emptied locally,
  unrelated to this Dockerfile) — already fixed locally, root cause
  still being looked into separately, not caused by and not blocking
  this PR.

## Question before this is ready for full review
 Does the toolchain-kept approach match what you had in mind here?


karttikjangid:
image
Follow line exercise is working .
@javizqh  can you also please do your testing and tell me if you find any problem i am also doing testing only

karttikjangid:
In the visual object detection it is showing up that the connecting to Robotics Backend for 10-15 mins without any error in logs , can you check from your side ?

PR #3918, opened 2026-07-09
n robots: one robot config per start pose

anishk85:
## n robots: one robot config per start pose

makes the world config carry N robots instead of always one, so the manager can
spawn a whole group. pairs with the multi-robot support in RoboticsApplicationManager.

### what changed
`scene.start_pose` can now hold either a single pose `[x, y, z, r, p, yaw]` or a
list of poses. we build one entry in `robot` per pose:

- a normal single-robot scene is just a list of one, so nothing special to handle
- the robot type is shared across the group
- each robot gets its own `entity` name (`<entity>_0`, `<entity>_1`, ... when
  there's more than one)

the entity is what the robot's launch file uses as its ros/gazebo namespace, so
N robots don't collide on topics, and gazebo can spawn and remove them one by one.

### DIFF
--- a/academy/views.py
+++ b/academy/views.py
@@ -476,17 +476,33 @@ def get_docker_world_data(fal, request):
             tools_configuration = json.loads(world.scene.tools_config)
 
         if world.robot.name != "None":
-            robot_config = [
-                {
-                    "name": world.robot.name,
-                    "launch_file_path": world.robot.launch_file_path,
-                    "ros_version": world.scene.ros_version,
-                    "type": world.scene.type,
-                    "start_pose": world.scene.start_pose,
-                    "entity": world.robot.entity,
-                    "extra_config": world.robot.extra_config,
-                }
-            ]
+            # one robot per start pose. scene.start_pose is either a single pose
+            # [x, y, z, r, p, yaw] or a list of them, so a normal single-robot
+            # scene is just a list of one and needs no special case. the robot
+            # type is shared; each robot gets its own entity name, which its
+            # launch file uses as the ros/gazebo namespace so N robots don't
+            # collide on topics, and which lets gazebo spawn and remove them
+            # one by one.
+            poses = world.scene.start_pose or []
+            if poses and not isinstance(poses[0], (list, tuple)):
+                poses = [poses]
+
+            robot_config = []
+            for index, pose in enumerate(poses):
+                entity = world.robot.entity
+                if len(poses) > 1:
+                    entity = f"{entity}_{index}"
+                robot_config.append(
+                    {
+                        "name": world.robot.name,
+                        "launch_file_path": world.robot.launch_file_path,
+                        "ros_version": world.scene.ros_version,
+                        "type": world.scene.type,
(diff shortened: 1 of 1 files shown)


anishk85:
didn't have much to change i don't know how the same changes were there in upstream.

academy/views.py
```diff
@@ -476,17 +476,33 @@ def get_docker_world_data(fal, request):
...
-                }
-            ]
+            # one robot per start pose. scene.start_pose is either a single pose
+            # [x, y, z, r, p, yaw] or a list of them, so a normal single-robot
+            # scene is just a list of one and needs no special case. the robot
+            # type is shared; each robot gets its own entity name, which its
+            # launch file uses as the ros/gazebo namespace so N robots don't
+            # collide on topics, and which lets gazebo spawn and remove them
+            # one by one.
+            poses = world.scene.start_pose or []
```
javizqh, line 486: Do not use or. If this variable does not have any data it **must** crash, because something really wrong has happened

academy/views.py
```diff
@@ -476,17 +476,33 @@ def get_docker_world_data(fal, request):
...
+            if poses and not isinstance(poses[0], (list, tuple)):
+                poses = [poses]
+
+            robot_config = []
+            for index, pose in enumerate(poses):
+                entity = world.robot.entity
+                if len(poses) > 1:
+                    entity = f"{entity}_{index}"
```
javizqh, line 494: This change must be done inside the RAM, not here

javizqh (changes requested):
This approach is completely wrong. If a world has 2 or more spawnpoints that does not mean that it has the same amount of robots.

The approach has to be to iterate through the robots and stop at the number of spawnpoints

anishk85:
got it, that makes sense — spawnpoints are just the available slots, not a robot count.

one thing i want to confirm before i change it: right now `World.robot` is a
OneToOneField, so a world can only hold a single robot — there's nothing to iterate
yet (though the docstring already says "one or more robots").

so to do this properly, should i change `World.robot` into a many-relation (e.g.
ManyToMany to Robot) + migration, so a world can hold N robots and i can iterate them
against the spawnpoints? that also means different robots per world, not N copies of
the same one.

and just to confirm on the entity: views should pass `robot.entity` as-is and RAM
makes it unique per instance — correct?

javizqh:
I will be working on that after the new version is released

anishk85:
okk so for the n robots parts i will work as per my local setup then.i have no choice later a lots of conflicts will arrive.

anishk85:
or if you want in my local setup i can change the db relation which i talked about how should i go ?

javizqh:
In this repository after the changes I will do to the databases next week, we will no longer need this PR because it will handle N robots natively. All other changes done here other than handling N robots are already wrong as explained in the last review

anishk85:
Okk so we should close this ig.

javizqh:
Already fixed in https://github.com/JdeRobot/RoboticsAcademy/pull/3927

PR #3938, opened 2026-07-22
run every entrypouint as its own agent

anishk85:


anishk85:
this pr refers to the excersice specific codes for drone cat and mouse

anishk85:
image
as per the previous architecture it was modified to support properly drone0 drone 1 namespaces instead of process a and process b 


given in the blog 
https://theroboticsclub.github.io/gsoc2026-Anish_Kumar/blog/2026/multiprocess-code-execution/

react_frontend/src/components/buttons/PlayPause.tsx
javizqh, line 1: This is not allowed. Please remove the changes to this file. This thing must be handled using the tags and not hardcoding It to the play button

javizqh (changes requested):

javizqh:
The Python changes are fine, but the others must be removed

react_frontend/src/components/buttons/PlayPause.tsx
anishk85: okk will look into the other way of solving this.

anishk85: hi so tags should carry the data or via porps tag->prop i should resolve this?

javizqh: Already fixed in https://github.com/JdeRobot/RoboticsAcademy/pull/3940

javizqh (approved):

PR #3940, opened 2026-07-23
Add db tag and write component tag

javizqh:
Add multiple entrypoint tag


anishk85:
thanks for helping with this

PR #3954, opened 2026-08-03
Add difficulty levels to drone cat mouse

anishk85:
[hard.webm](https://github.com/user-attachments/assets/2562c99b-2bc8-4d41-a9cc-492cc751fd60)
this was of hard level similarly other levels have been tested.

also tested drone rescue people.


anishk85:
video reference was attached above

javizqh:
Seems Ok, but remember for the next times to not break the order in the database

PR #3973, opened 2026-08-20
fix(compose): support Docker Desktop WSL2 NVIDIA workflow

karttikjangid:
## Summary

This PR updates the Windows-specific NVIDIA Docker Compose configurations to
support Docker Desktop with WSL2.

## Problem

The existing Windows NVIDIA Compose configuration attempts to expose `/dev/dri`.
That device is not available in the WSL2 GPU environment used by Docker Desktop,
where the GPU is exposed through the WSL2 graphics stack instead.

This causes the container to fail before the RoboticsAcademy application starts.

## Changes

- Updated `compose_cfg/dev_humble_nvidia_windows.yaml`.
- Updated `compose_cfg/user_humble_nvidia_windows.yaml`.
- Removed the `/dev/dri` device mapping from the Windows-specific NVIDIA
  configurations only.
- Corrected the duplicate `environment` definition in the Windows user
  configuration so the YAML has a single effective environment block.
- Kept the Linux and other platform Compose configurations unchanged.

## Validation

Tested on Windows 11 with Docker Desktop and WSL2:

- Docker Compose stack starts without the `/dev/dri` device error.
- NVIDIA GPU is visible inside the container.
- WSL2 GPU rendering was verified through the D3D12/Mesa path.
- The RoboticsAcademy container starts successfully.

## Scope

This PR is limited to the Docker Desktop + WSL2 Windows workflow.

It does not modify RoboticsApplicationManager, exercise code, or the Linux
Docker workflow.

### DIFF
--- a/compose_cfg/dev_humble_nvidia_windows.yaml
+++ b/compose_cfg/dev_humble_nvidia_windows.yaml
@@ -36,8 +36,7 @@ services:
                 - gpu
     environment:
       - LD_LIBRARY_PATH=/usr/lib/wsl/lib
-    devices:
-      - /dev/dri
+      - PYTHONPATH=/RoboticsApplicationManager
     ports:
       - "7164:7164"
       - "7163:7163"
--- a/compose_cfg/user_humble_nvidia_windows.yaml
+++ b/compose_cfg/user_humble_nvidia_windows.yaml
@@ -28,10 +28,6 @@ services:
     environment:
       - LD_LIBRARY_PATH=/usr/lib/wsl/lib
     runtime: nvidia
-    devices:
-      - /dev/dri
-    environment:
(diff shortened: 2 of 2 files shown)


compose_cfg/dev_humble_nvidia_windows.yaml
```diff
@@ -36,8 +36,7 @@ services:
                 - gpu
     environment:
       - LD_LIBRARY_PATH=/usr/lib/wsl/lib
-    devices:
-      - /dev/dri
+      - PYTHONPATH=/RoboticsApplicationManager
```
javizqh, line 39: This should not be needed as it has worked before without it

javizqh (changes requested):

Issue #3468, opened 2026-02-12
[Betatesting] Autoparking exercise using the 3d lidar


javizqh:
This is only for the new universe that has a 3d lidar, not the old one with 3 lasers. The documentation available is for the 3 lasers exercises, not the 3d lidar one.

Share here any issues, not solutions, related to this exercises. Do not comment anything else other than that.

This issue is not meant to be assigned to a single person, as its purpose is that multiple people test it.

JitinSaxenaa:
During testing of the Autoparking exercise in the new 3D LiDAR universe, I observed that the WebGUI LiDAR visualization appears slightly desynchronized from the Gazebo simulation.

Additionally, the parked vehicles along the roadside are not always clearly distinguishable in the LiDAR visualization. 

link for demo video - https://youtu.be/Z2dk8huQcjE

javizqh:
That is normal. We reduced the number of points shown, because a lot of them cause desync, but the drawback is that it shows less of the lidar

JitinSaxenaa:
Hi @javizqh 

Understood, thank you for the clarification.

Palakgokhru:
Issues I noticed:
When the car turns near parked vehicles, the LiDAR points in the WebGUI sometimes appear and disappear quickly.
While reversing, the obstacle points are not always shown continuously.
The WebGUI view looks slightly delayed compared to the Gazebo simulation, especially during sharper turns.

sriramsowmithri9807:
@javizqh assign me a issue to work on!!

Sourane09:
I've submitted PR #3726 adding HAL docstrings to address the missing 3D LiDAR documentation noted here.

javizqh:
The test period has ended

PR #3975, opened 2026-08-26
docs(autoparking): document current 3D LiDAR API

timothyanderson096-ocdealcheck:
## Summary

* Document the Python `HAL.getLidarData()` API.
* Replace six obsolete `/prius_autoparking/...` topics with the current `/autonomous_car/...` interfaces.
* Document the 3D LiDAR fields returned to users.

## Verification

The baseline documentation failed seven contract checks: six current ROS 2 topics and the Python LiDAR method were missing.

The corrected page passes all seven checks:

```text
PASS
  all six current ROS 2 topics are documented
  Python HAL.getLidarData() is documented
```

`git diff --check` also passes.

Addresses #3468.

This PR is intentionally documentation-only and does not claim completion of the full visual parking manoeuvre.


javizqh:
I will merge it just before a new version is released as the old documentation is still in use

PR #3985, opened 2026-09-02
Visual lander exercise update

aquintan4:


database/exercises/db.sql
```diff
@@ -209,7 +210,7 @@ COPY public.exercises_worlds (id, exercise_id, world_id, is_default) FROM stdin;
 79	31	78	True
 80	28	79	False
 81	28	80	False
-82	32	81	True
```
javizqh, line 212: Leave the 82 entry as it is and add entry 83

javizqh (changes requested):

javizqh (approved):

PR #3992, opened 2026-09-03
Offline RADI build scripts

aquintan4:


javizqh (changes requested):
Use the same dockerfile structure as the online one, just replace the git clones for COPY

